PR #1, opened 2022-03-25
Add Prepare input workflow

mahesh-panchal:
Adds a workflow to prepare channel input from a yaml file.

## PR checklist

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
  - [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/genomeassembler/tree/master/.github/CONTRIBUTING.md)
  - [ ] If necessary, also make a PR on the nf-core/genomeassembler _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [ ] Make sure your code lints (`nf-core lint`).
- [ ] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [ ] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).


fmalmeida (approved):
Have you put it to test? Just to make sure it is loading channels as expected?
I mean, after reading it, it looks very well structured/designed for me how it's parsed and channeled, just needing to make sure it runs as expected 😄 

I have just on comment, when using paired reads in the TSV, users would always have to pass the file pattern as a glob, right, as `my_pairs_{1,2}.fastq`.

subworkflows/local/prepare_input.nf
```diff
@@ -7,15 +7,16 @@ include { SAMTOOLS_FASTQ } from "../../modules/nf-core/modules/samtools/fastq/ma
...
 
     Simple Format:
 
-    ```csv
-    sample_id,data_type,paths
-    sample_A,HiFi,/path/to/hifi
+    ```tsv
+    sample_id   data_type   sequences
+    sample_A    HiFi   /path/to/hifi/reads.bam
+    sample_A    Illumina    /path/to/illumina/reads{1,2}.fastq.gz
```
d4straub, line 19: I have the same concern, whether `/path/to/illumina/reads{1,2}.fastq.gz` is the only way to specify paired end reads. It just assumes a certain naming scheme. I am not sure how that would work in the yml, but I assume multiple files can be specified?

With the current implementation, would it be possible to also use a list entry for PE illumina in the TSV? E.g.
```tsv
sample_id   data_type   sequences
sample_A    Illumina    ["/path/to/illumina/reads1.fastq.gz","/path/to/illumina/reads2.fastq.gz"]
```
If yes, would be great to have documented that.

mahesh-panchal:
> Have you put it to test? Just to make sure it is loading channels as expected? I mean, after reading it, it looks very well structured/designed for me how it's parsed and channeled, just needing to make sure it runs as expected 😄
> 
I haven't tested this one yet, but that can be taken care of in a sec. It's adapted from my own assembly workflow so there should only be a few syntax errors at most. I wanted feedback on the input, and if it's flexible enough but simple enough for people to work with.

> I have just on comment, when using paired reads in the TSV, users would always have to pass the file pattern as a glob as `my_pairs_{1,2}.fastq` ?

Yes, that field will take any file path including with globs, so I think pairs are best supplied in this format. Is there someone I'm not aware of that should be accounted for here? In the end, because this field can take any glob, it will be up to us to provide any extra checks and filtering to put it into the correct channels, e.g. separating single end from paired end.

subworkflows/local/prepare_input.nf
mahesh-panchal: I think `file` only accepts a valid string format for a path ( that can include globs). I don't think an array is possible like so. I was also hoping we could just use the cardinality to specify paired end vs single end as files belonging to the same sample will be grouped on exit. Is there something you had in mind about the specification of the reads?

d4straub: Yes I think you are right that it wont work as above.

> Is there something you had in mind about the specification of the reads?

I liked the sample sheet input in tsv / csv typically to specify forward and reverse file names, without globs or such. But I can accept the way you did it.
For PE it would be nice to maybe allowing comma separated entries or any other way specifying two paths. But maybe thats just me, than its not worth adding ;)

mahesh-panchal: Would you prefer the notation of `forward` and `reverse` instead of `read1`/`read2`?

d4straub: Hm, read1/read2 seems more generic. forward/reverse implicates a direction, e.g. stranded lib prep, amplicon seq. But the usual genome seq isn't stranded, so read1/read2 might be better? But the notation is to me not too important.

priyanka-surana:
What about 10x data?

mahesh-panchal:
Wasn't planning on incorporating 10X data as it's no long produced.

mahesh-panchal:
> Need to mark which inputs are for scaffolding and which are for assembly, or can this be determined from the file? E.g. Ultra long ONT

Can deal with this later. Let's just get a start on the rest of the modules.

priyanka-surana:
Looks good to me.

conf/test_input_yml.config
```diff
@@ -0,0 +1,27 @@
+/*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    Nextflow config file for running minimal tests
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    Defines input files and everything required to run a fast and simple pipeline test.
+
+    Use as follows:
+        nextflow run nf-core/genomeassembler -profile test,<docker/singularity> --outdir <OUTDIR>
```
d4straub, line 8:
```suggestion
        nextflow run nf-core/genomeassembler -profile test_input_yml,<docker/singularity> --outdir <OUTDIR>
```

d4straub (approved):
LGTM!
Very clean way to convert that files to channels. I should follow that example with ampliseq one day!

conf/test_input_yml.config
```diff
@@ -0,0 +1,27 @@
...
+
+----------------------------------------------------------------------------------------
+*/
+
+params {
+    config_profile_name        = 'Test YML input profile'
+    config_profile_description = 'Minimal test dataset to check pipeline function'
```
mahesh-panchal, line 15:
```suggestion
    config_profile_name        = 'Test YML input profile'
    config_profile_description = 'Minimal test dataset to check yaml input function'
```

conf/test_input_csv.config
```diff
@@ -0,0 +1,27 @@
+/*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    Nextflow config file for running minimal tests
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    Defines input files and everything required to run a fast and simple pipeline test.
+
+    Use as follows:
+        nextflow run nf-core/genomeassembler -profile test,<docker/singularity> --outdir <OUTDIR>
```
mahesh-panchal, line 8:
```suggestion
        nextflow run nf-core/genomeassembler -profile test_input_csv,<docker/singularity> --outdir <OUTDIR>
```

PR #56, opened 2024-04-30
ONT assembly

nschan:
I propose to add genome assembly from ONT or PacBio reads into this repo, based on a slack discussion with @mahesh-panchal . 
I am expecting that there are some changes that need to be made to make this nf-core compliant, but I am not sure which.


modules/busco/functions.nf
```diff
@@ -0,0 +1,59 @@
+/*
```
mahesh-panchal, line 1: Ooofff this is a blast from the past. Please use the updated modules.

main.nf
```diff
@@ -8,56 +8,85 @@
     Slack  : https://nfcore.slack.com/channels/genomeassembler
 ----------------------------------------------------------------------------------------
 */
-
-nextflow.enable.dsl = 2
+#!/usr/bin/env nextflow
```
mahesh-panchal, line 11: Given this is from an old template, you might be better off making a fresh pipeline with the new template.

modules/align/main.nf
```diff
@@ -0,0 +1,75 @@
...
+                                        publish_id:meta) }
+    input:
+        tuple val(meta), path(reads), path(reference)
+
+    output:
+        tuple val(meta), path("*.sam"), emit: alignment
+
+    script:
+        """
+        minimap2 -t $task.cpus \\
```
mahesh-panchal, line 23: You need to use the minimap modules here for this, and then you can combine this into a subworkflow if you want.

modules/genomescope/main.nf
```diff
@@ -0,0 +1,36 @@
+include { initOptions; saveFiles; getSoftwareName } from './functions'
+
+params.options = [:]
+options        = initOptions(params.options)
+
+process GENOMESCOPE {
```
mahesh-panchal, line 6: I think you can use what's integrated into the workflow already for this part.

modules/local/liftoff/main.nf
```diff
@@ -0,0 +1,46 @@
+include { initOptions; saveFiles; getSoftwareName } from './functions'
+
+params.options = [:]
+options        = initOptions(params.options)
+
+process LIFTOFF {
```
mahesh-panchal, line 6: This has a module on nf-core too, although you may want to integrate the tool Lifton instead

modules/local/ragtag/main.nf
```diff
@@ -0,0 +1,46 @@
+include { initOptions; saveFiles; getSoftwareName } from './functions'
+
+params.options = [:]
+options        = initOptions(params.options)
+
+process RAGTAG_SCAFFOLD {
```
mahesh-panchal, line 6: It looks like you'll need to contribute a module for this.

modules/local/slr/main.nf
```diff
@@ -0,0 +1,41 @@
...
+  
+  output:
+      tuple val(meta), path("${meta}_slr_scaffold.fa"), emit: scaffolds
+      tuple val(meta), path("${meta}_slr_unique.fa"),   emit: unique_contigs
+      tuple val(meta), path("${meta}_slr_ambigous.fa"), emit: abigous_contigs
+  
+  script:
+      def prefix = task.ext.prefix ?: "${meta}"
+  """
+  bwa index ${assembly}
```
mahesh-panchal, line 30: This looks like it would be a local module, or split between two given the SLR tool.
As the tools look common to a mulled container, you might be asked to use that so conda is also available to use.

mahesh-panchal (commented):
Added the comments in the Slack thread here too.

modules/align/main.nf
nschan: Ok, I will do that.

modules/local/liftoff/main.nf
nschan: Thanks for the suggestion, I will probably switch to lifton.

modules/local/slr/main.nf
nschan: This module has been deprecated since the tool caused problems and two long-read scaffolders and reference based scaffolder seem to be sufficient..

nschan:
Closing this and replacing with #62

PR #62, opened 2024-10-17
Assembler

nschan:
Hello,

this is a rather large update to the previous ONT assembly pipeline, namely that it now also does hifi assembly and combinations of ONT and HIFI.
There are additional improvements, such as kmer based QC with short reads and overall the pipeline has been refactored a bit.
This is a draft PR, because I am not yet using nf-core tooling, but I think that this could be added. However, I would like to have some discussion on this (mainly if it is acceptable for nf-core) before I invest time into restructuring everything around nf-core tooling.

Thanks for any comments!
Niklas

## PR checklist (will be updated once this moves from draft)

- [ x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/genomeassembler/tree/master/.github/CONTRIBUTING.md)
- [x] If necessary, also make a PR on the nf-core/genomeassembler _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [x] Make sure your code lints (`nf-core lint`).
- [x] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [x] Usage Documentation in `docs/usage.md` is updated.
- [x] Output Documentation in `docs/output.md` is updated.
- [x] `CHANGELOG.md` is updated.
- [x] `README.md` is updated (including new tool citations and authors/contributors).


mahesh-panchal:
Absolutely. Let's discuss. 

What specifically did you want to discuss? 
What's your target with this part of the pipeline? 
What's the scope of what you're going for? Something I wonder personally is whether to split this up a bit more given the other nf-core workflows. Contamination can be examined with #taxprofiler for example. This can guide the user whether then to use this pipeline to assemble, or whether to go for #mag. 

Assembly wise I still think Merqury is one of the most informative plots an assembly can have and should be a part of the pipeline. Given also the presence of sanger-tol's pipelines (I'm specifically thinking treeval and blobtools which would definitely be follow up pipelines) I'm wondering again the scope but there should be a very basic assembly pipeline at least that goes up to scaffolding.

What was this pipeline used to assemble so far? There's also the progress of the VGP and DToL pipelines and such. How much of that is incorporated here.

mahesh-panchal:
Also something really important though is do you have a test data set to help drive development? It's going to stall heavily if there's no test to help with integration work.

nschan:
> What specifically did you want to discuss?

Largely if this would be a useful addition to nf-core and if not what could make it one, you probably raise good points below.

> What's your target with this part of the pipeline?

The overall goal is genome assembly from **long reads**: ONT, HiFi or combinations of both. 
Generally the strategy is:

Only a single type of reads:
HiFi -> assembly with hifiasm
ONT -> assembly with flye
Alternatively: Assembly of HiFi with flye

ONT and HiFi reads:
Either: assemble individually (HiFi with hifiasm, ONT with flye) and then use HiFI assembly as a scaffold for the ONT assembly.
Or: use `hifiasm --ul` to combine both into a single assembly.

I built this pipeline to create genome assemblies of Arabidopsis thaliana ecotypes / strains, since this is _the_ plant model organisms there are good references available and this pipeline can work with or without references to guide assembly / scaffolding. 
I anticipate that sequencing of accessions / populations / strains / genotypes / (anything with genetic differences to a reference genome) with long-reads will soon be a standard for eukaryotes will small-mid genomes, given how cheap it got, which is why I think supporting the use of reference genomes is reasonable. 

> What's the scope of what you're going for? Something I wonder personally is whether to split this up a bit more given the other nf-core workflows. Contamination can be examined with #taxprofiler for example. This can guide the user whether then to use this pipeline to assemble, or whether to go for #mag.

This pipeline is intended for eukaryotic genome sequencing projects, probably starting from +/- pure DNA; metagenomic assembly is not within the scope of this pipeline. The idea is to go from (raw) reads to assembly and scaffolding of the assembly in a single pipeline. 
I think #mag is intended to assemble from metagenomic reads, Prokaryotes are probably covered by bacass? 

In case annotations are of interest: this pipeline can do liftoff from a reference. Generally, I think annotation is task which I think should be kept separate. I have an annotation pipeline set-up with some plant specific steps that is here https://github.com/nschan/nf-annotate 

> Assembly wise I still think Merqury is one of the most informative plots an assembly can have and should be a part of the pipeline. Given also the presence of sanger-tol's pipelines (I'm specifically thinking treeval and blobtools which would definitely be follow up pipelines) I'm wondering again the scope but there should be a very basic assembly pipeline at least that goes up to scaffolding.

The pipeline includes basic quality control using QUAST and BUSCO and runs yak (https://github.com/lh3/yak) to compare k-mer spectra between short-read datasets (if provided) and assemblies / scaffolded assemblies. I think the addition of meryl/merqury would be easily feasible. It may be worth fully swapping out yak for meryl/ -qury in that case.

> What was this pipeline used to assemble so far? There's also the progress of the VGP and DToL pipelines and such. How much of that is incorporated here.

I have used this pipeline to assemble A. thaliana genomes from ONT and/or HiFi reads; I am currently running tests to compare different assembly strategies. Regarding the other pipelines, I honestly do not know and would need to look into what exactly those pipelines do.

> Also something really important though is do you have a test data set to help drive development? It's going to stall heavily if there's no test to help with integration work.

I have ONT and HiFI  data (+ there are public short-reads) for four A. thaliana accessions, including the commonly used reference strain (Col-0). I could provide the Col-0 data as a test dataset.

mahesh-panchal:
It would definitely be useful to have something that does this in nf-core. 

My biggest issue is that I've not taken any time to make a test dataset partly because I really don't know what properties some of these test datasets should have. 

Does your test dataset run from start to end in the pipeline in less than 10 minutes?

nschan:
> My biggest issue is that I've not taken any time to make a test dataset partly because I really don't know what properties some of these test datasets should have.
I think this a difficult problems, ideally the test dataset should assembly at least one chromosome I would think. Creating something like that may be possible, via assembly -> mapping -> subsetting of the original reads? I have not looked into this,  and I assume that this would still not complete in in < 10 minutes; maybe something like human X or Y would be an option? I do not have any data for that though, I do plant science..

> Does your test dataset run from start to end in the pipeline in less than 10 minutes?
This is a full genome, so no. It takes around 10 hours on my infrastructure for a rather basic assembly that does not do all scaffolding steps.

mahesh-panchal:
Can you make something that's cut down? Not just a single chr, but even a small 2MB region or something.

nschan:
I probably could, yeah. 
I am currently testing if all combinations of parameters work for my full dataset. Once I have run all tests on the full datasets I have I should have everything (alignments) to create a smaller dataset from the reads.

nschan:
I have now added meryl / merqury to the pipeline. I also realised that what I have currently commited was not very up-to-date. Here is actual repo for the pipeline I would propose to add as `nf-core/genomeassembler` https://github.com/nschan/nf-arassembly .
Test data will come soon.

nschan:
I have not run tests with the test data so far, but I have added it here:
https://github.com/nschan/nf-genomeassembly/tree/master/assets/test_data

nschan:
So, if I understood you correctly, this could go into nf-core. I will then start setting things up with nf-core. I have a bunch of questions about this:

- Are there conventions around local subworkflows? Can I just use my local subworkflows and swap in nf-core modules without contributing the subworkflows to nf-core/modules/subworkflows ?
- Processes that have no modules in nf-core: Should I put those modules into local, or should I (try) to prepare a nf-core module? 
- Processes that have an nf-core module that I would like to change: Should this be turned into a local module? I understand that doing this is probably not the way it is intended and I will try to avoid doing it, but maybe it comes up.

Thanks
Niklas

mahesh-panchal:
Yes, although bear in mind things will change over time. There's definitely a need for duplicate purging, and the need for the potential of Hi-C curation, or at the very least the generate the initial maps. It would be nice if the input method I made could be kept, but it's not a huge problem if it's not. It's more important to get something working. 

Conventions are that yes, put everything in modules/local and subworkflows/local and then include nf-core modules as they're available. 

It's not a problem if the first draft is everything in local, but ideally try to get modules into nf-core modules and make recipes for bioconda ( within licensing constraints ). If there are things you want to change with an nf-core module, do make a PR. Keep in mind though that sometimes using nf-core modules might be really inefficient and so having a local module that pipes data between commands can quite happily live in modules/local. 

This kind of thing will all likely be discussed in the first review PR. Ideally though it should work on the test data set that is runnable on a desktop in 10-15 mins. 

As I see it the scope should go as far as to make an assembly ready for curation. Then other pipelines can take over.

https://pipelines.tol.sanger.ac.uk/ascc
https://pipelines.tol.sanger.ac.uk/treeval

I will also be trying to add what I have at some point: https://github.com/NBISweden/Earth-Biogenome-Project-pilot, run the hic branch with the test profile and you'll see the (partial) report.

nschan:
Ok, thanks. I will replace things with nf-core modules where available and add those into my subworkflows, I think that should be okay. I will open a new PR from a different branch for review once I have done that. Would you prefer to have a rather complete PR (i.e. including documentation, schema, etc) or would it be preferable to have the pipeline first and then add other things once the pipeline is mature?

nschan:
Hi,

I have decided to continue development in this branch. There are still some issues with the test dataset: it is large. The ONT subset is ~140mb, the pacbio reads are around 650mb. This does not work with nf-core/test-data I think. This is something that still needs to be figured out. For the next 7 days (or 100 downloads) the data can be downloaded here: https://syncandshare.lrz.de/getlink/fiGtuvVESRA2H1yqeGd1Y1/
The pipeline itself runs fine. There are still two things that I need to figure out: FASTQC and MULTIQC. I am not sure if fastqc is necessary, but MULTIQC would be really nice. I never used it before so it will probably take me a bit of time to figure out how to get everything in there, but I think the rest of the pipeline is ready for review. The samplesheet I used with the test profile looks like this:
```csv
sample,ontreads,hifireads,ref_fasta,ref_gff,shortread_F,shortread_R,paired
Col-0,ONT_Col-0_test_data.fastq.gz,Col_0.hifi_reads.fastq.gz,https://github.com/schatzlab/Col-CEN/raw/refs/heads/main/v1.2/Col-CEN_v1.2.fasta.gz,https://github.com/schatzlab/Col-CEN/raw/refs/heads/main/v1.2/Col-CEN_v1.2_genes.araport11.gff3.gz,not_tested.fastq.gz,,true
```

Thanks for reviewing this!
Niklas

mahesh-panchal:
I would only use FastQC for fixed length reads really, so Hi-C when it comes. Leave it out if your workflow doesn't need it.
As for MultiQC, that's a tricky one. It doesn't support most of the tools we're using yet. This means plugins would need to be written in python to parse tool logs. However, it's not a requirement. If you've checked out the pipeline I've been working on, you'll see I use MultiQC where I can, but it's all in a Quarto notebook which collects the results together in a report. A Quarto report or other such notebook is also suitable, but for future simplicity, I hope you go with Quarto.

nschan:
> I would only use FastQC for fixed length reads really, so Hi-C when it comes. Leave it out if your workflow doesn't need it.

The workflow can also take short genomic reads for polishing and / or k-mer QC, since these may just be available from a read archive. I guess fastqc could run on those.

> As for MultiQC, that's a tricky one. It doesn't support most of the tools we're using yet. This means plugins would need to be written in python to parse tool logs. However, it's not a requirement. If you've checked out the pipeline I've been working on, you'll see I use MultiQC where I can, but it's all in a Quarto notebook which collects the results together in a report. A Quarto report or other such notebook is also suitable, but for future simplicity, I hope you go with Quarto.

I will take a look at your quarto notebook approach. It would be _much_ easier for me to write some R into a Quarto notebook than trying to figure out MultiQC.

nschan:
Hello @mahesh-panchal, I so far completely ignored the failing tests assuming that I would deal with it later, but now that I took a look I have the impression that most of the test failures are not directly related to this pipeline, or at least it is not obvious to me what causes the test(s) to fail. If there is some configuration I need to do please let me know.

nschan:
I have added the tests with data from test-datasets.

mahesh-panchal:
My apologies, I'm a little busy this week to review. 

Linting you can fix by running `nf-core pipelines lint` ( I think that's the command now ), and it should tell you what to do. If you're running VScode, you may be able to select format document from a menu to help.

nschan:
No problem, I will try to finish the pipeline code today, the only thing that I am still struggling with is some formatting in the report; generating quarto documents dynamically turned out to be a bit more confusing than I had assumed. I will then try to deal with linting etc, but it seems the test-run now passes, which is good I guess. 
There will be some minor changes to some modules coming the next commits, but I think after that everything that is code should be ok. I guess what I am trying to say is, it probably makes sense to wait with reviewing until tomorrow / later when everything is ready..

nschan:
Here is the pipeline with reporting. The reporting script currently makes use of plotly to make plots fancier, but maybe this can be reconsidered for the `merqury` histograms. These contain some 10k datapoints each, which creates very large (50mb+) files, due to the plotly js conversion I assume.

Regarding the tests:
nf-core linting / nf-core (pull_request) fails during dependency installation:
```
 Successfully installed pip-24.3.1
ERROR: Invalid requirement: 'nf-core==': Expected end or semicolon (after name and no valid version specifier)
    nf-core==
```
I am not really sure where this is specified.

prettier fails with exit 1, I am also not sure why.

There is an additional issue with `nf-core pipelines lint`, which I have posted about in the slack https://nfcore.slack.com/archives/CE6PELWR4/p1730389681209279 and I currently do not really know what the problem or solution is, maybe you have run into this before?

mashehu:
@nf-core-bot fix linting

nschan:
@nf-core-bot fix linting

nschan:
@nf-core-bot fix linting

nschan:
@nf-core-bot fix linting

nextflow_schema.json
```diff
@@ -289,45 +204,185 @@
...
+            "type": "string",
+            "description": "file containing pacbio primers for trimming with lima"
+        },
+        "trim_short_reads": {
+            "type": "boolean",
+            "description": "trim short reads with trimgalore"
+        },
+        "assembler": {
+            "type": "string",
+            "description": "Assembler to use. Valid choices are: `'hifiasm'`, `'flye'`, or `'flye_on_hifiasm'`. `flye_on_hifiasm` will scaffold flye assembly (ont) on hifiasm (hifi) assembly using ragtag"
```
mashehu, line 273: use an enum to control the values of this param

nextflow_schema.json
```diff
@@ -289,45 +204,185 @@
...
+        },
+        "kmer_length": {
+            "type": "integer",
+            "description": "kmer length to be used for jellyfish"
```
mashehu, line 277: is there a min and a max value?

nschan: I assume that there is a range that makes sense, but I think it is nice to give people an option to try weird things?

nextflow_schema.json
```diff
@@ -289,45 +204,185 @@
...
+        },
+        "read_length": {
+            "type": "string",
```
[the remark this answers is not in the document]
nschan: No, it is an integer, thanks for pointing this out

conf/modules.config
d4straub, line 1: This file seems surprisingly empty, usually the output folder structure is defined here. And any additional "args" for modules to make customization with config files possible.

conf/test.config
```diff
@@ -10,20 +10,23 @@
...
-    max_time   = '6.h'
-
-    // Input data
-    input          = "$projectDir/assets/samplesheets/test_input_channels.yml"
-    mash_screen_db = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/sarscov2/genome/genome.fasta'
-
-    outdir     = 'results'
-    tracedir   = "${params.outdir}/pipeline_info"
-    publish_dir_mode = 'symlink'
+    input = "$projectDir/assets/samplesheets/test_samplesheet.csv"
```
d4straub, line 24: That might be fine for now, but later that should be rather in nf-core testdata repo

configs/ont_flye.config
d4straub, line 1: These configs seem fine, but with only 3 settings I am not sure they are necessary. Only two parameters would need to be set to get to that outcome, right?

docs.md
d4straub, line 1: The nf-core way would be mostly files docs/output.md & docs/usage.md & nextflow_schema.json

docs/output.md
d4straub, line 1: that needs amendment

docs/usage.md
d4straub, line 1: Ideally here would go most of the docu

d4straub (commented):
Only went though 62 of 355 files, but docs seem to need some updates ;)

nschan:
Thanks a lot for the suggestions @d4straub, I have not really worked on the docs so far, but I will try to get the documentation in a more useful state this week and will incorporate your comments. 
I see you point about the `ont_flye` profile, this is probably the one that is most easy to configure. I just thought it would make sense to have a complete set of configs, instead of only providing some for the "complex" cases and assume people can figure out the simpler ones on their own?

d4straub:
> I see you point about the ont_flye profile, this is probably the one that is most easy to configure. I just thought it would make sense to have a complete set of configs, instead of only providing some for the "complex" cases and assume people can figure out the simpler ones on their own?

Yes, sure, whatever makes the pipeline easier to use. When other cases are more complicated then it makes sense to make profiles for all (or common applications), I agree.

d4straub:
Its fine that documentation comes last! But `conf/modules.config` is not documentation and looks very different than what I am used to, e.g. https://github.com/nf-core/mag/blob/master/conf/modules.config or any other nf-core pipeline.

nschan:
> Its fine that documentation comes last! But conf/modules.config is not documentation and looks very different than what I am used to, e.g. https://github.com/nf-core/mag/blob/master/conf/modules.config or any other nf-core pipeline.

Thanks for pointing that out. It is a file I have not looked into yet, and will go onto my list for this week with a higher priority than the docs. I was not really aware of this file in nf-core, it explains a couple of things that were unclear to me with regards to the usage of nf-core modules. I will try to make use of it.

nschan:
@nf-core-bot fix linting

conf/modules.config
```diff
@@ -17,173 +17,91 @@ process {
...
-            mode: params.publish_dir_mode,
-            saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
+    /* 
+    ----------
+    ASSEMBLY
+    ----------
+    */
+    // FLYE: Here I can switch to nf-core/flye
+    withName: FLYE {
+        ext.ags = [
```
d4straub, line 27:
```suggestion
        ext.args = [
```

conf/modules.config
```diff
@@ -17,173 +17,91 @@ process {
...
+    publishDir = [
+        path: { "${params.outdir}/assembly/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" },
+        mode: params.publish_dir_mode,
+        saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
         ]
     }
-    withName: 'MERQURYFK_PLOIDYPLOT' {
+    // HIFIASM:
+    /* Don't think I should switch to nf-core 
+       here since the UL process takes more inputs */
```
d4straub, line 41: Some inputs are optional and using [] will omit them, depending on the module. Its also possible to use the nf-core module and modify it slightly and then [patch the module](https://nf-co.re/docs/nf-core-tools/modules/patch). Having said that, local modules are imho also fine.

conf/modules.config
```diff
@@ -17,173 +17,91 @@ process {
...
-    //         path: { "$params.outdir/library_comparison/kat_comp" },
-    //         mode: params.publish_dir_mode,
-    //         saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
-    //     ]
-    // }
-
-    // CONTAMINATION SCREEN
-    withName: 'MASH_SCREEN' {
-        ext.args   = '-w'
-        ext.prefix = { "${meta.id}_${query.baseName}" }
```
d4straub, line 131: These outer curly brackets here have a specific function: it executes the code only at runtime and by that making sure that params are parsed at runtime and not during config reading. That is optimal. You seem to not use that. This trick is missing in many configs of other pipelines (historic reasons), but if you do it new, why not optimal ;)

conf/modules.config
```diff
@@ -17,173 +17,91 @@ process {
...
-        cpus       = { Math.min(6, consensus instanceof List ? consensus.size() : 1 ) }
+    /* 
+    ----------
+    Polishing
+    ----------
+    */
+    // Medaka: switch to nf-core
+    withName: MEDAKA {
+        ext.args = [ params.medaka_model ? "--model ${params.medaka_model}" : '' ]
+        ext.prefix = '"${meta.id}_medaka"'
```
d4straub, line 91: I hdont recall the use of ' around args, why is that used here? I'd rather propose
```suggestion
        ext.prefix = { "${meta.id}_medaka" }
```

conf/modules.config
```diff
@@ -17,173 +17,91 @@ process {
...
         publishDir = [
-            path: { "$params.outdir/assembly_comparison/quast" },
+            path: { "${params.outdir}/polishing/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" },
```
d4straub, line 93: The `${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}` works I guess but for readability its better to write the folder name diretely. But might be just a matter of preference.

d4straub (commented):
The modules.config looks quite good I think, nice!

conf/modules.config
nschan: Ah, great, thanks for the explanation, I was wondering how the behaviour is controlled while writing the config and arrived at using `''` to try to control when things are substituted, I will use `{}` (see also your comment below)

conf/modules.config
nschan: Good point, I will make things explicit unless I need to dynamically set one of the folders.

nschan:
@nf-core-bot fix linting

conf/igenomes_ignored.config
mahesh-panchal, line 1: Just delete this file and the related channels.

conf/modules.config
```diff
@@ -17,173 +17,296 @@ process {
...
 
-    // COMPARE_ASSEMBLIES
-    withName: 'QUAST' {
-        cpus       = { Math.min(6, consensus instanceof List ? consensus.size() : 1 ) }
+    // BUSCO: nf-core
+    withName: BUSCO {
+        ext.prefix = { "${meta.id}-${task.process.tokenize(':')[-3].toLowerCase()}-${lineage}" }
         publishDir = [
-            path: { "$params.outdir/assembly_comparison/quast" },
+            path: { "${params.outdir}/QC/${task.process.tokenize(':')[2].toLowerCase()}${task.process.tokenize(':')[2].toLowerCase() == 'assemble' ? '' : '/' + task.process.tokenize(':')[-3].tokenize('_')[1].toLowerCase() }/busco/${meta.id}" },
```
mahesh-panchal, line 182: Why is this complicated expression needed to form the directory name?

conf/modules.config
```diff
@@ -17,173 +17,296 @@ process {
...
-        tag        = { meta.build }
-        ext.prefix = { meta.build }
+    /* 
+    ----------
+    Polishing
+    ----------
+    */
+    // Medaka: local; nf-core module cant deal with gzipped input
+    withName: MEDAKA {
+        ext.args = [params.medaka_model ? "--model ${params.medaka_model}" : '' ]
```
mahesh-panchal, line 232: A general guideline is that any `ext.*` directive should be a closure if it uses a variable. https://nf-co.re/docs/guidelines/components/modules#optional-command-arguments

mahesh-panchal (commented):
Here are some starting comments. I'll keep going slowly when I have time.

workflows/genomeassembler.nf
```diff
@@ -1,345 +1,288 @@
...
+    Channel.empty().set { ch_hifi_reads }
+    Channel.empty().set { ch_shortreads }
+    Channel.empty().set { meryl_kmers }
+    Channel.empty().set { ch_flye_inputs }
+    Channel.empty().set { ch_hifiasm_inputs }
+    Channel.empty().set { genome_size }
+    Channel.empty().set { ch_versions }
+
+    Channel
+        .fromPath("$projectDir/assets/report/.DUMMYFILE") // quast emits a single file
```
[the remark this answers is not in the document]
nschan: This file is to have a file input when the process (e.g. QUAST) was not part of the pipeline run. I followed the nextflow patterns for optional input https://nextflow-io.github.io/patterns/optional-input/

workflows/genomeassembler.nf
```diff
@@ -1,345 +1,288 @@
...
+    Scaffolding
+    */
+    
+    SCAFFOLD(ch_input, ch_longreads, ch_polished_genome, ch_refs, ch_ref_bam, meryl_kmers)
+
+    /*
+    Report
+    */
+        
+    quast_files
```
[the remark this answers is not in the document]
nschan: I admit I do not find the suggested style very nice, but I will run the formatter to have the pipeline consistent with the recommended style.

workflows/genomeassembler.nf
```diff
@@ -1,345 +1,288 @@
...
+    Channel.empty().set { ch_shortreads }
+    Channel.empty().set { meryl_kmers }
+    Channel.empty().set { ch_flye_inputs }
+    Channel.empty().set { ch_hifiasm_inputs }
+    Channel.empty().set { genome_size }
+    Channel.empty().set { ch_versions }
+
+    Channel
+        .fromPath("$projectDir/assets/report/.DUMMYFILE") // quast emits a single file
+        .tap { quast_files }
```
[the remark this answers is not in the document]
nschan: Do you mean I should / could use `set` ?

assets/report/_quarto.yml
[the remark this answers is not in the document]
nschan: That file actually contained a mistake (it should only be `defaults: base`, the one with all profiles in defaults is what I used for testing..).
The reason why this I did it like that is quarto related, and I hope that there is a better way to do it that I missed. Here is how I understand quarto does conditional reports:
I can pass params to quarto via `-P name:value` and profiles via `--profile profilename`. I need params to do conditional execution of R code chunks via `#| eval: !expr params$name`. At the same time, I need to use profiles to make use of quartos ability to toggle the visibility of divs, which can be done with `::: { .content-visible unless-profile="name" }` and `::: { .content-visible when-profile="name"}`, but there seems to be no `unless-param` or `when-param` for divs. There would be a `-meta` but that also does not help because the R chunks can only access `params`. What this boils down to is: I need profiles for divs and I need params for code chunks. I am also not able to set params via the profile files, which would make things a bit easier.

workflows/genomeassembler.nf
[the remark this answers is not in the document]
mahesh-panchal: This pattern has a weakness in that if a process has two or more optional inputs, and you try this, you'll get a file collision. If you pass an empty list, then it just works. It's what we've done in nf-core for a long-time now.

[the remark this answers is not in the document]
nschan: Alright, I have updated this.

conf/modules.config
```diff
@@ -17,171 +17,294 @@ process {
...
-    // CONTAMINATION SCREEN
-    withName: 'MASH_SCREEN' {
-        ext.args   = '-w'
-        ext.prefix = { "${meta.id}_${query.baseName}" }
+    // QUAST: Prefer to keep the local module since it can deal with the inputs I have
+    withName: QUAST {
+        ext.prefix = { "${meta.id}_${task.process.tokenize(':')[-3].toLowerCase()}" }
         publishDir = [
-            path: { "$params.outdir/contamination_screen/mash_screen/screens" },
+            path: { "${params.outdir}/QC/${task.process.tokenize(':')[2].toLowerCase()}${task.process.tokenize(':')[2].toLowerCase() == 'assemble' ? '' : '/' + task.process.tokenize(':')[-3].tokenize('_')[1].toLowerCase()}/quast/${meta.id}" },
```
mahesh-panchal, line 192: Is there a necessity for this complex expression? Why not use `meta.id` all the time?

configs/hifi_flye.config
```diff
@@ -0,0 +1,6 @@
+params {
+    assembler = 'flye'
```
mahesh-panchal, line 2: I feel like I'm missing something with this set of profiles. Only one of these can be run at a time right? Should this profile just be a workflow parameter?

modules/local/bam_sort_stat/bam_stats_samtools.nf
```diff
@@ -0,0 +1,25 @@
...
+// From https://github.com/nf-core/viralrecon/blob/dev/subworkflows/nf-core/bam_stats_samtools.nf
+//
+
+params.options = [:]
+
+include { SAMTOOLS_STATS    } from '../samtools/stats/main'    
+include { SAMTOOLS_IDXSTATS } from '../samtools/idxstats/main' 
+include { SAMTOOLS_FLAGSTAT } from '../samtools/flagstat/main' 
+
+workflow BAM_STATS_SAMTOOLS {
```
mahesh-panchal, line 12: Doesn't nf-core have this module?

modules/local/genomescope/main.nf
```diff
@@ -0,0 +1,31 @@
+process GENOMESCOPE {
```
mahesh-panchal, line 1: Why not the nf-core module?

modules/local/hifiasm/main.nf
```diff
@@ -0,0 +1,80 @@
...
+    script:
+    def args = hifi_args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    hifiasm \\
+        $args \\
+        -l0 \\
+        -o ${prefix}.asm \\
+        -t $task.cpus \\
+        --ul ${ont_reads} \\
```
mahesh-panchal, line 74: You should make a PR to nf-core to add this extra functionality that's missing.

modules/local/quast/main.nf
mahesh-panchal, line 1: Why did you move this file?

modules/local/samtools/fastq/main.nf
```diff
@@ -0,0 +1,21 @@
...
+    tuple val(meta), path(bam), path(bai)
+
+    output:
+    tuple val(meta), path("*.fq.gz"), emit: fasta
+
+    script:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+    samtools fastq $bam | gzip | ${prefix}.fq.gz
```
mahesh-panchal, line 19: This must surely be throwing an error

modules/local/samtools/flagstat/main.nf
```diff
@@ -0,0 +1,23 @@
+process SAMTOOLS_FLAGSTAT {
```
mahesh-panchal, line 1: I'm a little confused why there are so many modules on nf-core that are local instead.

subworkflows/local/ont/main.nf
```diff
@@ -0,0 +1,41 @@
...
+
+workflow ONT {
+    take:
+    input_channel
+
+    main:
+    Channel.empty().set { genome_size }
+    Channel
+        .fromPath("${projectDir}/assets/report/.DUMMYFILE")
+        .map { it -> [null, it] }
```
mahesh-panchal, line 12: inserting `null`'s in a workflow is generally not advised. Either use empty lists, empty maps, or empty strings where appropriate. But also remember to remove these instances of `.DUMMYFILE` as nf-core has a better way of doing it.

subworkflows/local/polishing/pilon/run_pilon/main.nf
```diff
@@ -0,0 +1,20 @@
...
+  assembly_in
+  aln_to_assembly_bam_bai
+
+  main:
+  assembly_in
+    .join(aln_to_assembly_bam_bai)
+    .set { pilon_in }
+
+  PILON(
+    pilon_in.map { it -> [it[0], it[1]] },
```
mahesh-panchal, line 13: Ideally, so others can also follow, be more explicit with your variable names.
Avoid using `it`, and when closures have identifiable elements the split the input i.e. `{ meta, bam -> ... }`

subworkflows/local/prepare_hifi/main.nf
```diff
@@ -0,0 +1,23 @@
...
+workflow PREPARE_HIFI {
+  take:
+  inputs
+
+  main:
+  inputs
+    .map { it -> [it.meta, it.hifireads] }
+    .set { hifireads }
+  if (params.lima) {
+    if (params.pacbio_primers == null) {
```
mahesh-panchal, line 13:
```suggestion
    if (!params.pacbio_primers) {
```

modules/local/jellyfish/main.nf
```diff
@@ -0,0 +1,98 @@
...
+    input:
+        tuple val(meta), path(kmers)
+
+    output:
+        tuple val(meta), path("*.txt"), emit: stats
+
+    script:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+        """
+        jellyfish stats $kmers > ${prefix}_stats.txt
```
mahesh-panchal, line 71: Will you swap to FastK / Meryl, or are you planning on sticking with Jellyfish?

modules/local/genomescope/main.nf
nschan: The nf-core module does not emit what I need, the estimated length.

assets/report/report.qmd
```diff
@@ -0,0 +1,746 @@
+---
+title: "nf-core/genomeassembler report"
+author: "Niklas Schandry"
+format: dashboard
```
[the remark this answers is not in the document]
nschan: I think the main benefit is that it is more "clicky" and less "scrolly" than the report, mostly things go into tabs and not below each other.

bin/genomescope.R
```diff
@@ -0,0 +1,732 @@
+#!/usr/bin/env Rscript
+
+## GenomeScope: Fast Genome Analysis from Unassembled Short Reads
+## 
+## This is the automated script for computing genome characteristics 
+## from a kmer histogram file, k-mer size, and readlength
+
+## https://github.com/schatzlab/genomescope/
```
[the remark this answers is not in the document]
nschan: Yes, that is a good point. I will make use of the container used by the nf-core module for this (cp your other comment).

conf/modules.config
nschan: QUAST (and the other QC tools), will run after each of the steps in this assembly; i.e. `assembly`, `polish` and `scaffold`, and polish and scaffold may also be different tools. If I only used `meta.id` this would put every output in the same, which may overwrite things, and is a huge mess. This way they go into e.g. `polish/pilon/$meta.id` or `polish/medaka/$meta.id` for the two polishing options.

modules/local/samtools/fastq/main.nf
nschan: I will take a look when I go through the samtools module(s) again.

modules/local/samtools/flagstat/main.nf
nschan: See above; 
> Yes, I will see why I am using my local version (which is an older nf-core version) again. I remember trying to use the nf-core module and something did not work anymore, but I will take another look,

modules/local/jellyfish/main.nf
nschan: Meryl is used in the pipeline for merqury QC.
Jellyfish/genomescope part is _only_ to estimate genome size from ONT reads, which may be a) not necessary since flye no longer requires a genomesize and b) not very accurate since it involves some guessing about the read length. I am not sure if this is very important functionality.

conf/modules.config
mahesh-panchal: I think it's clearer to use subworkflows then and use a pattern match against the subworkflow to specify this.
```nextflow
process {
    withName: '.*:POLISH:QUAST' {
        publishDir = ...
    }
    withName: '.*:PILON:QUAST' {
        publishDir = ...
    }
```
It also makes it easier to add a tool option to a given stage of using quast.

configs/hifi_flye.config
mahesh-panchal: There's a toy example [here](https://github.com/mahesh-panchal/Nextflow-metadata-practices) of how to implement profiles for workflows. This way has the benefit that it could be extended to support running multiple profiles, unlike nextflow configuration profiles which would overwrite each other.

conf/modules.config
nschan: That is much clearer, I agree.

modules/local/jellyfish/main.nf
mahesh-panchal: Genomescope is very important QC check of the data. Since ONT reads are getting more accurate, it's definitely better to have this functionality, although Genomescope can be run from Meryl/FastK databases too. Also note, there's GOAT ( Genomes on a tree ) which can be queried to get species information estimates like genome size, haploid number, etc. It's a good sanity check that the assembly size doesn't vary wildly from the estimated genome size from genomescope or GOAT, but there are reasons it might.

configs/hifi_flye.config
nschan: I will look into this, it would be nice to be able to run multiple profiles at the same time.

modules/local/bam_sort_stat/bam_stats_samtools.nf
nschan: This should be resolved now.

modules/local/jellyfish/main.nf
nschan: OK, the I will leave this in. Do you think adding a check (in the pipeline) for this is necessary? It is currently possible to do this in the generated report, but this is left to the person running the pipeline. I think at some point the user has to actually do some sanity checking themselves and I do not know to which degree delegating this into the pipeline is going to make it easier?

subworkflows/local/polishing/pilon/run_pilon/main.nf
nschan: That is a good suggestion, I will take it into account in the future

configs/hifi_flye.config
nschan: I have taken a look at your example, and while I think I get it, adding this into the current pipeline would be a bit of extra effort, I would probably need to rewrite the assembly module and redesign some part of the input handling. Do you think its worth it at this stage? I am happy to continue working on this pipeline after it is released and would prefer to have the initial release happen sometime soonish and then add some nice-to-have feature in a later release?

mahesh-panchal (approved):
Approving without a full review so the first release PR can take place.

PR #65, opened 2025-01-16
First release

nschan:
This is a release PR to merge dev into master. I have accidentally already created main from dev, if main is preferred over master I can remove main and open a PR.

This pipeline was originally initiated by @mahesh-panchal I think. This is however a complete rewrite from scratch of the pipeline.

## PR checklist

- [ x] This comment contains a description of changes (with reason).
- [x ] If you've fixed a bug or added code that should be tested, add tests!
- [x ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/genomeassembler/tree/master/.github/CONTRIBUTING.md)
- [x ] If necessary, also make a PR on the nf-core/genomeassembler _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [ x] Make sure your code lints (`nf-core pipelines lint`).
- [x ] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [x ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ x] Usage Documentation in `docs/usage.md` is updated.
- [ x] Output Documentation in `docs/output.md` is updated.
- [x ] `CHANGELOG.md` is updated.
- [ x] `README.md` is updated (including new tool citations and authors/contributors).

PR #67, opened 2025-01-20
Second set of reviews

nschan:
This PR addresses some more of the reviews on #65:

- Some small changes to the reporting script; I also noticed that reporting did not run during tests that do not include `merqury`, and I have made changes to avoid that.
- switched to a patched nf-core `liftoff` module
- changed regex for gff3 files
- switch to nf-core hifiasm
- some typos


modules/nf-core/liftoff/liftoff.diff
LouisLeNezet, line 1: Why did you modify so much liftoff ?
I would recommend to only patch only the first channel of input to suit your necessity and keep everything else even if not necessary. This will make it easier to update this tool in case someone update it on nf-core.

LouisLeNezet (changes requested):
See my comment below.
I would recommend to patch just what's necessary.

modules/nf-core/liftoff/liftoff.diff
nschan: I made the patch smaller, but I would like to keep the functionality that detects if the input is gzipped or not in here.

nschan:
This PR now also contains updates to use the [updated nf-core hifiasm module](https://github.com/nf-core/modules/pull/7336)

modules/nf-core/liftoff/liftoff.diff
LouisLeNezet: Your addition could be a good improvment to the nf-core module.
You can add a issue to propose your change 😉

LouisLeNezet (approved):
Seems good to me !

PR #69, opened 2025-01-21
fix assemble subworkflow

nschan:
After switching to the updated hifiasm module in the assembly subworkflow, I added functionality to use `--ont` with hifiasm. This contained a mistake.


LouisLeNezet (approved):
I missed this one in the revious review !

PR #70, opened 2025-01-23
module configuration

nschan:
With the switch to the nf-core module I also need to update the configuration of that module.


nschan:
This PR has grown to also include configuration for hifiasm. There was also an update to the report to deal with the changed input names.

LouisLeNezet (approved):
Seems good to me !

PR #76, opened 2025-02-10
GHA: include conda in tests

nschan:
I noticed that GHA only runs with `docker`, I think it would make sense to also include `conda` and `singularity`.


nschan:
Singularity runs seem to not be possible on the runner(s), I have removed the singularity tests.

LouisLeNezet (approved):
Good catch

PR #77, opened 2025-02-10
use https container for singularity

nschan:
The report module used an oras container for singularity. Since I do not use singularity I was not aware that this does not work with older singuarlity versions. With this I switch to the https container.


LouisLeNezet (approved):
Seems fine for me

PR #78, opened 2025-02-10
README: switch metromap based on theme

nschan:
The metromap displayed in the readme now switches between light/dark version based on GH color scheme.


LouisLeNezet (approved):
That's a good idea 👍 
I'll add that to our pipeline !!

PR #79, opened 2025-02-10
GHA: update CI-tests

nschan:
I noticed that the tests were missing the assembler that was tested, now the tests are named `NXF_VER | assembler | profile`, instead of `NXF_VER | | profile`.

The `download_pipeline` action did not specify which assembler to test, it now tests `hifiont_flyehifiasm`, which runs both flye and hifiasm.

### DIFF
--- a/.github/workflows/download_pipeline.yml
+++ b/.github/workflows/download_pipeline.yml
@@ -110,7 +110,7 @@ jobs:
         env:
           NXF_SINGULARITY_CACHEDIR: ./singularity_container_images
           NXF_SINGULARITY_HOME_MOUNT: true
-        run: nextflow run ./${{ needs.configure.outputs.REPOTITLE_LOWERCASE }}/$( sed 's/\W/_/g' <<< ${{ needs.configure.outputs.REPO_BRANCH }}) -profile test,hifiont_flyehifiasm,singularity --outdir ./results
+        run: nextflow run ./${{ needs.configure.outputs.REPOTITLE_LOWERCASE }}/$( sed 's/\W/_/g' <<< ${{ needs.configure.outputs.REPO_BRANCH }}) -profile test,singularity --outdir ./results
 
       - name: Count the downloaded number of container images
         id: count_afterwards
--- a/.github/workflows/download_pipeline.yml
+++ b/.github/workflows/download_pipeline.yml
@@ -103,14 +103,14 @@ jobs:
         env:
           NXF_SINGULARITY_CACHEDIR: ./singularity_container_images
           NXF_SINGULARITY_HOME_MOUNT: true
-        run: nextflow run ./${{needs.configure.outputs.REPOTITLE_LOWERCASE }}/$( sed 's/\W/_/g' <<< ${{ needs.configure.outputs.REPO_BRANCH }}) -stub -profile test,singularity --outdir ./results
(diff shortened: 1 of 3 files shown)


.github/workflows/download_pipeline.yml
```diff
@@ -103,14 +103,14 @@ jobs:
         env:
           NXF_SINGULARITY_CACHEDIR: ./singularity_container_images
           NXF_SINGULARITY_HOME_MOUNT: true
-        run: nextflow run ./${{needs.configure.outputs.REPOTITLE_LOWERCASE }}/$( sed 's/\W/_/g' <<< ${{ needs.configure.outputs.REPO_BRANCH }}) -stub -profile test,singularity --outdir ./results
+        run: nextflow run ./${{needs.configure.outputs.REPOTITLE_LOWERCASE }}/$( sed 's/\W/_/g' <<< ${{ needs.configure.outputs.REPO_BRANCH }}) -stub -profile test,hifiont_flyehifiasm,singularity --outdir ./results
```
LouisLeNezet, line 106: Why do you need to modify this one ?
Shouldn't the `test` profile be enough ?

nschan: The `test` profile currently is not enough, since it does not come with a default assembler strategy and errors out (neither `params.ont` or `params.hifi` are set).  `test` only specifies resources and disables qc steps.  This is why all of the other CI tests use a combination of `test` and `matrix$ASSEMBLER` to run the tests.

nschan: The error message that currently comes out is: `error('At least one of params.ont, params.hifi needs to be true.')`, which is probably not extremely obvious. I can of course specify an assembler in the `test` profile. Am I correct in thinking that if i do `-profile test,other` `other` will overwrite things set by `test`?

LouisLeNezet: > I can of course specify an assembler in the test profile.
I think that would be better for new users. If they want afterwards to use this same profile but with another assembler they will be able to overwrite it through the command line.

> Am I correct in thinking that if i do -profile test,other other will overwrite things set by test?
I think it should, but double check it. Nextflow is always a bit tricky with argument priority.

nschan: According to the docs, it is: When more than one of these options for specifying configurations are used, they are merged, so that the settings in the first override the same settings appearing in the second, and so on.
I have therefore changed the order of for CI to `-profile ${{ matrix.ASSEMBLER }},test,${{matrix.profile}}` and made `test` identical to hifiont_flyehifiasm.

.github/workflows/download_pipeline.yml
```diff
@@ -110,7 +110,7 @@ jobs:
         env:
           NXF_SINGULARITY_CACHEDIR: ./singularity_container_images
           NXF_SINGULARITY_HOME_MOUNT: true
-        run: nextflow run ./${{ needs.configure.outputs.REPOTITLE_LOWERCASE }}/$( sed 's/\W/_/g' <<< ${{ needs.configure.outputs.REPO_BRANCH }}) -profile test,singularity --outdir ./results
+        run: nextflow run ./${{ needs.configure.outputs.REPOTITLE_LOWERCASE }}/$( sed 's/\W/_/g' <<< ${{ needs.configure.outputs.REPO_BRANCH }}) -profile test,hifiont_flyehifiasm,singularity --outdir ./results
```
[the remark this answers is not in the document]
nschan: No :( I only modified the stub run

LouisLeNezet (approved):
LGTM 😄

PR #82, opened 2025-02-11
Add stubs for local modules

nschan:
There is an issue with the [`download` GHA](https://github.com/nf-core/genomeassembler/actions/runs/13259363850).
This continues to fail during automated testing, there is no obvious error message and I am a bit confused what is going on. For now I have added stubs for the local modules that are part of this test, to see if I can get a successful stub run. While doing this, I realised that there are no stubs for any of the local modules. I will add those in the near future. 
I also noticed that test.config contained a duplicated resourceLimits block, this has been removed.

### DIFF
--- a/.github/workflows/download_pipeline.yml
+++ b/.github/workflows/download_pipeline.yml
@@ -10,7 +10,7 @@ on:
       testbranch:
         description: "The specific branch you wish to utilize for the test execution of nf-core pipelines download."
         required: true
-        default: "dev"
+        default: "gha_download"
   pull_request:
     types:
       - opened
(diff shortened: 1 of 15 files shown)


nschan:
I have now added stubs for all local modules.
While doing this I noticed that the `QUAST` module contained some older logic to determine the `prefix` that has meanwhile been put into `modules.conf` but not removed from the QUAST module. I have removed that part from the QUAST main.nf.

nschan:
This PR has not solved the issue with the downloads action, but it has lead to the creation of stubs for all local modules, which should probably go into dev.

PR #83, opened 2025-02-11
use different container for collect_reads

nschan:
The `collect_reads` module was using a python3.11 container, but it seems more reasonable to use the coreutils / grep / gzip container and environment that is also used by `gfa2fa`.


LouisLeNezet (approved):
Looks good to me

PR #84, opened 2025-02-11
Fix ci download

nschan:
An attempt to test the download GHA in a PR.


nschan:
This did not give the desired result.

PR #86, opened 2025-02-13
Longstitch outputs; report header

nschan:
The longstitch module sometimes (?) emitted two files in the output channels, since the original file was not removed after the one named based on `prefix` was created.

The report.qmd header contained extra indentation (cp #85 which I will close after merging this one).


LouisLeNezet (approved):
Seems good to me

PR #87, opened 2025-02-13
quast output path

nschan:
When updating quast, I forgot a `/` in the `results` output path.


nschan:
Correction: The problem was using `${prefix}` in the outputs, which is not defined. Using `${meta.id}*` should catch all prefixes.

LouisLeNezet (approved):

PR #96, opened 2025-02-14
reviews: remove duplicate from usage; add short info to configs

nschan:
more reviews from @mirpedrol


mirpedrol (approved):
thanks for the changes!

PR #97, opened 2025-02-14
update LIFTOFF patterns

nschan:
I noticed a mistake in the patterns for liftoff; since it is included in the `RUN_TOOL` subworkflows, the pattern must not include the `:` before `TOOL`


mirpedrol (approved):
LGTM

PR #99, opened 2025-02-17
remove multiqc settings

nschan:
This removes some multiqc leftovers from the template in nextflow.config and nextflow_schema.json

### DIFF
--- a/.nf-core.yml
+++ b/.nf-core.yml
@@ -18,3 +18,6 @@ template:
   org: nf-core
   outdir: .
   version: 1.0.0
+  skip_features:
+    - multiqc
+    - igenomes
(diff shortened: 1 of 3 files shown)


mirpedrol (approved):
You can also add the following to your `.nf-core.yml` file, this will avoid the template sync in following nf-core/tools updates to keep adding these lines to your pipeline. (last 3 lines)
```
template:
  author: Niklas Schandry
  description: Assemble genomes from long ONT or pacbio HiFi reads
  force: false
  is_nfcore: true
  name: genomeassembler
  org: nf-core
  outdir: .
  version: 1.0.0
  skip_features:
    - multiqc
    - igenomes
```

Once this is added LGTM :)

PR #100, opened 2025-02-18
Update module output paths, remove samtools::index, adjust report code, update scaffolding tool prefix handling

nschan:
This PR updates the module output paths to a hopefully easier to use structure, trying to put files that I assume users will expect to be in the same folder into the same folder. I am also trying to reduce the overall number of output folders and depth of the tree.
While doing this, I noticed that nf-core minimap2 includes an indexing step. This lead to duplicated index files for the same mapping file. I have removed the dedicated indexing step from the pipeline to avoid this. 

There are some commits in here that were part of #99, I had to rebase while working on this.

I have also updated the functions parsing QC from different tools in the report to match the changed names. The suffix for initial assemblies changed from `assemble` to `assembly`.

While working through this and revisiting all modules, I also updated how the scaffolding tools [links, longstitch, ragtag] handle the `ext.prefix` to make it consistent with other modules, and also more useful.


modules/local/ragtag/main.nf
```diff
@@ -10,39 +10,40 @@ process RAGTAG_SCAFFOLD {
     tuple val(meta), path(assembly), path(reference)
 
     output:
-    tuple val(meta), path("${assembly}_ragtag_${reference}/*.fasta"), emit: corrected_assembly
-    tuple val(meta), path("${assembly}_ragtag_${reference}/*.agp"), emit: corrected_agp
-    tuple val(meta), path("${assembly}_ragtag_${reference}/*.stats"), emit: corrected_stats
+    tuple val(meta), path("*.fasta"), emit: corrected_assembly
+    tuple val(meta), path("*.agp"), emit: corrected_agp
+    tuple val(meta), path("*.stats"), emit: corrected_stats
```
mirpedrol, line 15: this could be aligned, it makes it easier to read
```suggestion
    tuple val(meta), path("*.fasta"), emit: corrected_assembly
    tuple val(meta), path("*.agp"),   emit: corrected_agp
    tuple val(meta), path("*.stats"), emit: corrected_stats
```

mirpedrol (approved):
LGTM, not that I can't review the science behind these changes as I don't know the pipeline, only the code standards

PR #104, opened 2025-02-24
Template -> 3.2.0

nschan:
This is a template sync to 3.2.0, which seems to not have synced properly before

### DIFF
--- a/workflows/genomeassembler.nf
+++ b/workflows/genomeassembler.nf
@@ -227,7 +227,7 @@ workflow GENOMEASSEMBLER {
     softwareVersionsToYAML(ch_versions)
         .collectFile(
             storeDir: "${params.outdir}/pipeline_info",
-            name: 'nf_core_' + 'pipeline_software_' + 'versions.yml',
+            name: 'nf_core_'  +  'genomeassembler_software_'  + 'versions.yml',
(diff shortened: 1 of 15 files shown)


workflows/genomeassembler.nf
```diff
@@ -227,7 +227,7 @@ workflow GENOMEASSEMBLER {
     softwareVersionsToYAML(ch_versions)
         .collectFile(
             storeDir: "${params.outdir}/pipeline_info",
-            name: 'nf_core_' + 'pipeline_software_' + 'versions.yml',
+            name: 'nf_core_'  +  'genomeassembler_software_'  + 'versions.yml',
```
LouisLeNezet, line 230: Is it something that was modified in the TEMPLATE or is it something that you've done ?

nschan: This happened with the template, I was also a bit surprised but it is actually useful for me since I use `nf_core_pipeline_software_versions.yml` to report versions in the report and this file here then also contains the versions that were used to create the report :)

LouisLeNezet: I will need to check that on ours 😄

nschan:
What should I do about the failing singularity tests?

PR #105, opened 2025-02-24
Fix singularity CI, clean nextflow.config

nschan:
This PR changes:

`ci.yml`: I have added back in the installation step(s) for singularity and conda

`nextflow.config`: When revisiting this file for singularity setup I noticed that a lot of the profiles contained duplicated blocks, I have removed those.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,8 +85,8 @@ jobs:
           echo $(realpath $CONDA)/condabin >> $GITHUB_PATH
           echo $(realpath python) >> $GITHUB_PATH
 
-      # - name: Clean up Disk space
-      #   uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be # v1.3.1
+      - name: Clean up Disk space
+        uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be # v1.3.1
 
       - name: "Run pipeline with test data ${{ matrix.ASSEMBLER }}"
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Install Nextflow
+      - name: Set up Nextflow
         uses: nf-core/setup-nextflow@v2
         with:
           version: "${{ matrix.NXF_VER }}"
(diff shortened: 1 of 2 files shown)


nschan:
While this PR was supposed to fix the failing tests with singularity, it instead seems to have broken all other tests. The tests get killed (exitcode 137), which I assume is out of memory. I don't why this happens, the test configs were not changed and worked fine before.

nschan:
I have disabled the clean diskspace step, now all tests pass.

nschan:
I have re-enabled clean disk space, confirming it leads to subsequent tests getting killed with 137. I am now checking if the clean up step can be run after testing the pipeline.

.github/workflows/ci.yml
```diff
@@ -36,19 +36,58 @@ jobs:
...
           - "hifiont_hifiasm"
           - "hifiont_flyehifiasm"
-        profile: ["conda", "docker", "singularity"]
+        profile:
+          - "conda"
+          - "docker"
+          - "singularity"
+        isMaster:
+          - ${{ github.base_ref == 'master' }}
+        # Exclude conda and singularity on dev
```
LouisLeNezet, line 45: Why did you comment those lines ?
This might take a while to compute each time you have a PR on dev.

nextflow.config
LouisLeNezet, line 1: Good catch !

.github/workflows/ci.yml
nschan: I will uncomment those lines, I commented them because I assumed otherwise the tests would not run in this PR against dev.

mirpedrol (approved):
I don't know why cleaning the disk space made the tests fail in your case, this step was added to solve some space problems we found in other pipelines. But since your tests are passing without this step, it's ok to remove it 👍

.github/workflows/ci.yml
```diff
@@ -1,5 +1,5 @@
 name: nf-core CI
-# This workflow runs the pipeline with the minimal test dataset to check that it completes without any syntax errors
+# {% raw %}This workflow runs the pipeline with the minimal test dataset to check that it completes without any syntax errors
```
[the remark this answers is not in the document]
mirpedrol: yes, sorry, I wrote this before you cahged it 🙂

mirpedrol:
> Just to clarify, should I remove the clean disk space step, or run it after the test? I assume it doesn't hurt to clean up the runner after the test finished?

It doesn't hurt to clean after the test, but I doubt it has an effect since the runners are shut down and a new runner is created when needed. But I think it's ok to leave it as it is for now 👍

PR #109, opened 2025-02-27
clean nextflow.config, remove empty channels, reorganize jellyfish

nschan:
This PR addresses changes proposed by @LouisLeNezet during review:

- Cleaned up nextflow.config to remove duplicated lines and empty `process{}`
- Removed empty channels from the `HIFI` subworkflow
- Renaming of emitted outputs from `ASSEMBLE` in `emit` block
- Broke `jellyfish` into several files, one per process (earlier comment), adjusted `JELLYFISH` subworkflow includes


nschan:
This is a preliminary PR for the latest reviews. I am not sure if jellyfish `STATS` and samtools `STATS` should be removed from the pipeline. Personally, I would prefer to keep those `STATS` in the pipeline since I think they produce outputs that can be useful, even if they do not go into report.

LouisLeNezet (approved):
Everything is fine for me

LouisLeNezet:
> This is a preliminary PR for the latest reviews. I am not sure if jellyfish STATS and samtools STATS should be removed from the pipeline. Personally, I would prefer to keep those STATS in the pipeline since I think they produce outputs that can be useful, even if they do not go into report.

I agree, my comment was more about informing the user that this file exist.

nschan:
Good point about mentioning that these files exist. The bam stats are listed in `output.md`,but jellyfish + genomescope do not show up at all. I will add them!

PR #115, opened 2025-02-28
patch merqury for AWS test

nschan:
This is another attempt at getting merqury to work with the test dataset.
We have updated the short-read test data in https://github.com/nschan/test-datasets/tree/genomeassembler to have more differences between shortread and assembled sequences. Still, this dataset remains artificial and only produces a single contig/sequence during assembly, which means that some of the merqury files are not created.


nschan:
If this also does not solve the problems with the merqury test (may create problems during report?), I think it would be better to disable merqury QC for the aws test, unless that is generally not recommended.

PR #116, opened 2025-02-28
patch merqury for AWS test

nschan:
Another attempt to make outputs optional to get through with the full_test data


mirpedrol (approved):
🤞

PR #117, opened 2025-03-03
AWS-test: busco output dirs

nschan:
This is an attempt to fix the failing AWS megatest. It fails because BUSCO emits the lineage-dir into `QC/BUSCO` and it seems that this causes problems with overwriting in the outdir when there are subfolders emitted that are not empty (at least that is my understanding of the error message). For this test only, the busco steps now emit into a subworkflow specific subdir (`QC/BUSCO/<stage>`). This error is specific for runs where the lineage files are downloaded by the pipeline, and probably S3. I think users should pre-fetch the busco lineage for production runs, otherwise each BUSCO process will download the lineage files again.

### DIFF
--- a/conf/test_full.config
+++ b/conf/test_full.config
@@ -11,53 +11,35 @@
 */
 
 process {
-withName: '.*:ASSEMBLE:.*:BUSCO' {
-        ext.prefix = { "${meta.id}_assembly-${lineage}" }
+    withName: '.*:ASSEMBLE:.*:BUSCO' {
         publishDir = [
             path: { "${params.outdir}/${meta.id}/QC/BUSCO/assemble" },
-            mode: params.publish_dir_mode,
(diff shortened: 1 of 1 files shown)


conf/test_full.config
```diff
@@ -10,6 +10,59 @@
 ----------------------------------------------------------------------------------------
 */
 
+process {
+withName: '.*:ASSEMBLE:.*:BUSCO' {
+        ext.prefix = { "${meta.id}_assembly-${lineage}" }
+        publishDir = [
+            path: { "${params.outdir}/${meta.id}/QC/BUSCO/assemble" },
+            mode: params.publish_dir_mode,
+            saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
```
mirpedrol, line 19: I think you don't need to override `ext.prefix` or the `publishDir` `mode` and `saveAs`, as they are the same in `modules.config`

mirpedrol (approved):

PR #125, opened 2025-03-10
use correct genome-size for flye and longstitch

### LINKED ISSUE #124: Hardcoded -G in longstitch
The value of -G in longstitch is hardcoded but should be based on provided or estimated genome-size.

nschan:
The `longstitch` module contained a hardcoded genome-size value, an oversight that should not have made it into the release. 
While fixing this, I realized that the way `genome_size` was handled with `flye` was also bad, since it relied on changing `params.genome_size`, if there was an estimated sample-specific value from genomescope; I am not even sure that worked correctly. Generally, now the `genome_size` channel is used and either contains the value of `params.genome-size`, which can be updated with the estimated genome-size from `genomescope` in the `ONT` module.

For flye, passing genome_size is now handled by _temporarily_ storing genome_size as part of the `meta`-map, from where can be used with `ext.args` via `modules.conf`, to avoid having to patch the `flye` module. Of course, this has to be removed from `meta` afterwards to avoid breaking all later joins. 

For `longstitch`, since this is a local module, I have added genome_size to the inputs.

This closes #124.

### DIFF
--- a/subworkflows/local/assemble/main.nf
+++ b/subworkflows/local/assemble/main.nf
@@ -16,7 +16,7 @@ workflow ASSEMBLE {
     ont_reads // meta, reads
     hifi_reads // meta, reads
     ch_input
-    genomescope_out
+    genome_size
     meryl_kmers
 
     main:
@@ -57,13 +57,14 @@ workflow ASSEMBLE {
             }
             if (params.ont) {
                 ont_reads.set { flye_inputs }
-                if (params.genome_size == null && params.jellyfish) {
-                    params.genome_size = genomescope_out
-                }
             }
             // Run flye
+            flye_inputs
+                .join(genome_size)
+                .map { meta, reads, genomesize -> [[id: meta.id, genome_size: genomesize], reads] }
+                .set { flye_inputs }
             FLYE(flye_inputs, params.flye_mode)
-            FLYE.out.fasta.set { ch_assembly }
+            FLYE.out.fasta.map { meta, assembly -> [[id: meta.id], assembly] }.set { ch_assembly }
             ch_versions = ch_versions.mix(FLYE.out.versions)
         }
         if (params.assembler == "hifiasm") {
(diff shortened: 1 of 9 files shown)


workflows/genomeassembler.nf
```diff
@@ -82,12 +82,14 @@ workflow GENOMEASSEMBLER {
...
+    ch_input.map { it -> [it.meta, params.genome_size] }
+            .set { genome_size }
 
     /*
     ONT reads
     */
     if (params.ont) {
-        ONT(ch_input)
+        ONT(ch_input, genome_size)
         ONT.out.genome_size.set { genome_size }
```
mirpedrol, line 93: is this line needed now?

subworkflows/local/assemble/main.nf
```diff
@@ -57,13 +57,14 @@ workflow ASSEMBLE {
...
             if (params.ont) {
                 ont_reads.set { flye_inputs }
-                if (params.genome_size == null && params.jellyfish) {
-                    params.genome_size = genomescope_out
-                }
             }
             // Run flye
+            flye_inputs
+                .join(genome_size)
+                .map { meta, reads, genomesize -> [[id: meta.id, genome_size: genomesize], reads] }
```
mirpedrol, line 64: just a suggestion, depends on your code style, but in case you wanted to keep the meta map as is and add a new key, or remove one key in the line below, instead of writing all the meta keys (you only have `id` in this case but it might get more complicated in the future if you add more keys), you can do this

```
meta - meta.subMap('genome_size') 
meta + [ genome_size: genomesize ]
```

mirpedrol (approved):
only one small comment, otherwise LGTM

workflows/genomeassembler.nf
nschan: This line is mainly needed to pass the input genomesize through the channel if it is set, i.e. be able to emit the unchanged channel from ONT. If `params.genome_size` is set that value will be used, if it is unset it will be estimated per-sample in the ONT workflow, but I think the user-provided value should take priority.

subworkflows/local/assemble/main.nf
nschan: Oh, thats neat, I think that is nicer than what I am doing, I will update my code

workflows/genomeassembler.nf
mirpedrol: okn thanks for the explanation.

mirpedrol (approved):

Issue #118, opened 2025-03-04
Troubles with genomeassembler workflow


BrunoBertolasi:
Hello to all, 

I am new to nf and genomeassembler and I am having some difficult to understand errors and causes of process failures. I have installed and tested successfully the genomeassembler workflow. I then tried to use genomeassembler for a full genome assembly of my real data, but the process stopped after having completed only two tasks.

My PC configuration includes a 15 core (11th Gen Intel(R) Core(TM) i7-11700K @ 3.60GHz) with 128 Gb of RAM. 
The starting file, containing the reads to assemble in format .gz, is 12.9 Gb.
I used this command:
```
nextflow run nf-core/genomeassembler \
 -profile singularity \
 -c /home/bruno/Desktop/test.config \
 --input /home/bruno/Desktop/samplesheet.csv \
 --outdir /home/bruno/Desktop/Acer_opalus_assembly \
 --ont true \
 --porechop true \
 --hifi false \
 --short_reads false \
 --polish_pilon false \
 --polish_medaka true \
 --lift_annotations false
```

The test.config file contains the line:
 `singularity.pullTimeout = "888h"`
  
The workflow started correctly but, after about two hours, I got this message:
```
-[nf-core/genomeassembler] Pipeline completed with errors-
ERROR ~ Execution aborted due to an unexpected error

 -- Check script '.nextflow/assets/nf-core/genomeassembler/./workflows/../subworkflows/local/scaffolding/./ragtag/../../mapping/map_to_assembly/../../../../modules/nf-core/minimap2/align/main.nf' at line: 7 or see '.nextflow.log' file for more details
ERROR ~ Pipeline failed. Please refer to troubleshooting docs: https://nf-co.re/docs/usage/troubleshooting

 -- Check '.nextflow.log' file for details
```

For the sake of precision I report the lines of the logfile in which an error appears:
 
 ```
Mar-03 17:49:55.033 [Task monitor] DEBUG n.processor.TaskPollingMonitor - !! executor local > tasks to be completed: 1 -- submitted tasks are shown below
~> TaskHandler[id: 1; name: NFCORE_GENOMEASSEMBLER:GENOMEASSEMBLER:ONT:PREPARE_ONT:CHOP:PORECHOP (Acer_opalus_full); status: RUNNING; exit: -; error: -; workDir: /home/bruno/work/3b/5be87ce8f051b93db2498503c3284a]
[...]

PR #126, opened 2025-03-17
Jellyfish singularity container

nschan:
as reported in #118 there is still a wrong singularity url for some of the jellyfish processes. These are corrected here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,9 @@ Bugfix release
 
 ### `Fixed`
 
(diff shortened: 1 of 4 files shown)


fellen31 (commented):
Looks good to me when the checks are passing!

maxulysse:
Can you push anything (even an empty commit) just to retrigger the tests?

nschan:
That failing test is new and I do not really see how it would relate to the changes in this PR, not sure what causes it.

maxulysse (approved):
CI error is unrelated

PR #130, opened 2025-03-19
version bump 1.1.0dev

nschan:
Bumping version of dev branch post release

### DIFF
--- a/.nf-core.yml
+++ b/.nf-core.yml
@@ -20,4 +20,4 @@ template:
   skip_features:
     - multiqc
     - igenomes
-  version: 1.0.1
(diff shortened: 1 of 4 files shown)


jasmezz:
What's the error with the double naming of the 'CHOP' workflow that we see in the failing CI checks? I saw it was also present for some tests of the release PR #128 ... It seems a bit worrying because here no pipeline test is succeeding.

nschan:
This is a problem in the current edge-release (https://github.com/nextflow-io/nextflow/pull/5894), it is not specific for this pipeline. The tests with 24.04.2 pass.

PR #131, opened 2025-03-24
refactor qc into subworkflow

nschan:
This PR refactors all QC into a subworkflow.

This should make future maintenance and extension of QC easier, since changes to QC steps are all happening in a single place.


d4straub (approved):
LGTM!

Issue #135, opened 2025-03-31
add ragtag `patch` to `flye_on_hifiasm`


nschan:
### Description of feature

Currently `flye_on_hifiasm` uses only ragtag for scaffolding. It would probably be good to also add in ragtag `patch`, to replace Ns with sequences.

nschan:
This was done in #136.

Issue #110, opened 2025-02-27
Contribute back to nf-core/modules


LouisLeNezet:
### Description of feature

There is a lot of local modules present in this pipeline.
It would be really nice to have them integrated into the nf-core/modules repository.

This could be part of a Hackathon 😉

nschan:
To make it easier, here is a list of local modules and my personal, subjective judgement on whether they should be in nf-core/modules, or if there is a module already that I did not use:

| Module | What does it do? | nf-core module candidate? |
| --- | --- | --- |
| `collect_reads` | `gunzip`s fastq.gz files into one fastq file, this useful because ONT can create many files per library if the run is restarted. etc | no |
| `genomescope` |  runs genomescope, most important for pipeline functionality is that this estimates the estimated haploid length, which gets passed on to `flye` | [a nf-core module exists ](https://github.com/nf-core/modules/blob/master/modules/nf-core/genomescope2/main.nf) but did not suit my needs |
| `gfa2fa` | converts `gfa` to `fa`, essentially `awk` | maybe? |
| `jellyfish` | runs jellyfish to create inputs for genomescope (see above), split into `COUNT`, `STATS`,`HISTO` and `DUMP` | yes |
| `links` | runs [LINKS](https://github.com/bcgsc/LINKS) for scaffolding | yes |
| `longstitch` | runs [longstitch](https://github.com/bcgsc/longstitch) for scaffolding | yes |
| `medaka` | runs [medaka](https://github.com/nanoporetech/medaka) for polishing. A [`medaka` nf-core module exists](https://github.com/nf-core/modules/tree/master/modules/nf-core/medaka), but runs an older version. My implementation here may need to be optimized, but also the [future of medaka is not clear and it might be good to switch to dorado](https://github.com/nanoporetech/medaka/issues/546) | probably / update nf-core module |
| `longstitch` | runs [longstitch](https://github.com/bcgsc/longstitch) for scaffolding | yes |
| `nanoq` | runs nanoq, [nf-core module exists](https://github.com/nf-core/modules/blob/master/modules/nf-core/nanoq/), but does not emit something I need (median read length) | module exists |
| `QUAST` | Has an nf-core module. The nf-core module uses a biocontainer that is missing something (I forgot what exactly, there was some discussion about this with @jfy133 on slack), this local module uses a different container | [module exists](https://github.com/nf-core/modules/tree/master/modules/nf-core/quast) |
[...]

PR #136, opened 2025-03-31
Ragtag patch for consensus assemblies

nschan:
Currently, `flye_on_hifiasm` uses ragtag `scaffold`, while it seems that ragtag `patch` would be more appropriate see #135 

Changes in this PR:

- Added `ragtag/patch` module
- moved `ragtag/scaffold` module into separate module file
- updated `assemble` subworkflow to use `ragtag/patch`
- updated `ragtag scaffold` subworkflow to use `ragtag/scaffold` from correct module path
- remove empty channels from `ragtag scaffold` subworkflow, those are handled in `QC` (missed those in #131)
- updated changelog


modules/local/ragtag/patch/main.nf
```diff
@@ -0,0 +1,75 @@
+process RAGTAG_PATCH {
```
jfy133, line 1: Is there a reason why this is a local module and not an official module?

jfy133: I would also recommend adding a `meta.yaml` and `main.nf.test` even if a local module :)

jfy133 (approved):
Looks fine to me for a local module, but could go further for completeness ;)

modules/local/ragtag/patch/main.nf
jfy133: > Not really, I think an nf-core ragtag module would be good, but requires time & effort (cp discussion in https://github.com/nf-core/genomeassembler/issues/110)

I would suggest you just do that now (I'm happy to do a rapid review), as you're almost there already to be honest and once you've got your `meta.yaml`, unless you have no experience with nf-test - you can just copy over the files you've already got and write a the nf-test file. 

Better to do that up front and pick away at it to reduce future pain (you may be increasingly blocked on future pipeline releases if you do not)

nschan: > unless you have no experience with nf-test - you can just copy over the files you've already got and write a the nf-test file.

I do not have any experience with nf-test. I would like to avoid future pain though

jfy133: Maybe this will help ;) 

https://nf-co.re/docs/tutorials/nf-core_training/writing-nf-core-modules/chapter-6-testing-your-module

nschan:
@jfy133, I have switched to the nf-core modules for this. Since that was your main comment on this PR, I assume that is fine and I can merge?

jfy133 (approved):

PR #138, opened 2025-04-09
update ragtag module

nschan:
Update the ragtag module after https://github.com/nf-core/modules/pull/8248 was merged.


famosab (approved):
LGTM

PR #145, opened 2025-04-14
check if assembly is skipped in initialization, closes #143

### LINKED ISSUE #143: `--skip-assembly` throws error if `--assembler` is `flye` and both ONT and HiFI reads are provided
### Description of the bug

Reported via slack https://nfcore.slack.com/archives/C038XMTQWBT/p1744381245036579?thread_ts=1743699666.061259&cid=C038XMTQWBT

nschan:
Checking if assemblies should actually be done before throwing an error regarding assemblies. See #143


LouisLeNezet (approved):
Seems also good to me

Issue #144, opened 2025-04-14
`hifiasm` based hybrid assemblies


nschan:
### Description of feature

Currently, there is only the option to scaffold ONT assemblies from `flye` on HiFi assemblies from `hifiasm`. However, `hifiasm --ont` performs very well, maybe even better that the `hifiasm --ul` algorithm. It might therefore be useful to be able to scaffold `hifiasm --ont` onto `hifiasm` (hifi reads) assemblies.

PR #146, opened 2025-04-14
add hifiasm_ont_hifiasm_on_hifiasm profile, code, test, docs

nschan:
This PR implements #144: using `hifiasm` for individual assemblies of ONT and HiFi, subsequently scaffolding ONT onto HiFi; the rationale is that generally it seems that hifiasm is better for repeats, ONT is better for most other parts.

To make this a bit less confusing for users, I have renamed the `hifiont_flyehifiasm` profile to `hifiont_flye_on_hifiasm` and added profile `hifiont_hifiasm_on_hifiasm`.

### DIFF
--- a/configs/hifi_ont_hifiasm_on_hifiasm.config
+++ b/configs/hifi_ont_hifiasm_on_hifiasm.config
@@ -1,0 +1,12 @@
+/*
+    Use this config to:
+        assemble HIFI reads with hifiasm
+        assemble ONT reads with flye
+        scaffold the flye assembly onto the hifiasm assembly
+*/
+
+params {
+    hifi = true
+    ont = true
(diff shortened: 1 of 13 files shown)


configs/hifi_ont_hifiasm_on_hifiasm.config
```diff
@@ -0,0 +1,12 @@
+/*
+    Use this config to:
+        assemble HIFI reads with hifiasm
+        assemble ONT reads with flye
+        scaffold the flye assembly onto the hifiasm assembly
```
jkh00, line 5: this should probably be:

assemble ONT reads with hifiasm 
scaffold the ONT assembly (hifiasm) onto HiFi assembly (hifiasm)

jkh00 (approved):

PR #148, opened 2025-04-14
switch to LINKS nf-core module

nschan:
As part of [module migration to nf-core](https://github.com/nf-core/genomeassembler/issues/110) LINKS has been [added into nf-core](https://github.com/nf-core/modules/pull/8260) and the local module is removed.


LouisLeNezet (approved):
Seems good to me !

PR #149, opened 2025-04-16
enable mapping to ref for hifiasm_on_hifiasm

nschan:
This fixes an oversight where the mapping to reference was skipped with `hifiasm_on_hifiasm`, because the reads were not specified for this strategy.


fellen31 (commented):
No changelog update?

nschan:
The changelog was updated when this option was added to dev in #146, 2 days ago. Since this does address a bug caught pre-release, I figured an extra update would not add much information, but I am ok with updating it if necessary.

fellen31 (approved):

Issue #152, opened 2025-04-29
Pipeline failing at QUAST


cchapus:
### Description of the bug

Dear all-

I'm running the nf-core/genomeassembler pipeline in order to assemble bacterial genome from ONT and Illumina reads (combining them).

At the quast step, my pipeline failed with the following message:
```
 /usr/local/bin/quast.py --output-dir A2_assembly -r ref.fasta --threads 12 A2.assembly.fasta.gz --glimmer --ref-bam A2_to_reference.bam --bam A2_assembly.bam --large
  
  Version: 5.2.0
  
  System information:
    OS: Linux-6.8.0-58-generic-x86_64-with-glibc2.28 (linux_64)
    Python version: 3.9.13
    CPUs number: 24
  
  Started: 2025-04-29 16:31:37
  
  Logging to A2_assembly/quast.log
  
  CWD: work/48/64ff86cbbfddde28a712c777263081
  Main parameters: 
    MODE: large, threads: 12, eukaryotic: true, min contig length: 3000, min alignment length: 500, \
    min alignment IDY: 95.0, ambiguity: one, min local misassembly length: 200, min extensive misassembly length: 7000
  
  Reference:
    ref.fasta ==> ref
  
  Contigs:
    Pre-processing...
    A2.assembly.fasta.gz ==> A2.assembly
  
  2025-04-29 16:31:38
  Running Reads analyzer...
  Compiling BWA (details are in /usr/local/lib/python3.9/site-packages/quast_libs/bwa/make.log and make.err)
  
  [Errno 30] Read-only file system: '/usr/local/lib/python3.9/site-packages/quast_libs/bwa/make.log'
  Traceback (most recent call last):
    File "/usr/local/bin/quast.py", line 309, in <module>
      return_code = main(sys.argv[1:])
    File "/usr/local/bin/quast.py", line 118, in main
      bed_fpath, cov_fpath, physical_cov_fpath = reads_analyzer.do(ref_fpath, contigs_fpaths,
    File "/usr/local/lib/python3.9/site-packages/quast_libs/reads_analyzer.py", line 928, in do
      if not compile_reads_analyzer_tools(logger):
    File "/usr/local/lib/python3.9/site-packages/quast_libs/ra_utils/misc.py", line 115, in compile_reads_analyzer_tools
      if compile_bwa(logger, only_clean) and compile_bedtools(logger, only_clean) and compile_minimap(logger, only_clean):
[...]

PR #161, opened 2025-05-19
update report containers

nschan:
Add singularity container


nschan:
This should fix the report issue reported in #152

jkh00 (approved):

PR #163, opened 2025-06-02
Report: singularity image link

nschan:
Another attempt at correctly specifying the singularity image. See #152


nschan:
I honestly do not really know what works, since I do not have setup that I can use to test singularity. It was reported in #152 that the container did not work, but now that I took another look at the error message that might actually not be the case

mashehu:
you should be able to test it with codespaces here on github

mashehu:
I think you are not using the correct https link (it is more than just changing the protocol tag in the beginning). See here in how to get it: https://nf-co.re/docs/tutorials/nf-core_components/using_seqera_containers

nschan:
You are probably correct about that, I did not realize that there is an https checkbox that only appears later; I have updated the link

Issue #160, opened 2025-05-19
Per-sample assembly strategy & params?


nschan:
As is the case with many pipelines, I wonder if it would be an improvement to switch from having the assembly-strategy as a pipeline parameter towards having this as a per-sample setting.
The advantage of this would be that it should then be easier to compare how different assemblers, etc perform on a dataset. For this to be actually useful, it would then also be necessary to pass sample-specific args to the assembly steps. 
The main argument against doing this is that would increase complexity for users, and it would require a bit of work in the pipeline; mainly updates to the assemble workflow and reporting as well as some overall refactoring to carry those sample-wise settings through the pipeline as part of the meta-map. 

As I currently imagine this, this would extend the samplesheet by these columns:
`ont_assembler, ont_assembler_args, hifi_assembler, hifi_assembler_args,scaffolding`
Where `ont_assembler` specifies the assembler for ONT reads, `hifi_assembler` specifies the assembler for hifi reads, `*_args` specifies tool specific args, and `scaffolding` indicates if one assembly should be scaffolded on the other (as e.g. flye_on_hifiasm) do, but may also be used to toggle hybrid assembly for assemblers that support it (e.g. `hifiasm --ul`). `scaffolding` could e.g. take values `ont_on_hifi`, `hifi_on_ont` and `none`, which should cover all current options.

nschan:
This idea has been implemented in #171 .

PR #171, opened 2025-07-09
v2.0.0: Refactor for sample-wise parameterisation

nschan:
_Updated Feb 04 2026_

As suggested [here](https://github.com/nf-core/genomeassembler/issues/160) this is full refactor of `genomeassembler` to support sample-level parameterisation of _everything_.

**Why?**
Often when doing genome assembly, we do not know what works best. With this change, this pipeline can be used to compare different settings for the same set of reads, to compare the assembly outcome. Samples that share the same value in `group` will be combined during reporting and preprocessing to facilitate comparisons of strategies on the same input(s). 

**Per-sample parameterisation**

Initially, I implemented this poorly, based largely around `join()`'ing various maps back to other maps. nextflow does not have a `join`operator for maps, so this was a big mess and turned out to be hard to read, annoying to write, and constantly blocking. To summarise: bad idea, cannot recommend. This attempt contributes to the large number of commits in this branch.

While contemplating my failure to implement sample-level parameterisation, I realised that the solution to this problem is "`meta`-stuffing", also referred to as "`meta`-smuggling" by @prototaxites, who seems to have arrived at a similar conclusion at around the same time.
In the purest form, this would implement everything in a single `meta`-map, which is in slot `0` of the channel traveling through the pipeline. I will use `meta` to refer to `[0]` of list-channels from here on.
Note: A pure `meta`-stuffing implementation would have required additional refactoring of some subworkflows, in particular of `QC` which takes more than one input channel, something I did not want to do.

_How this works:_
Everything goes into `meta`: `params` are turned into k/v pairs for each sample, unless the samplesheet contains a different value for that sample with the same key, resulting in a massive `meta`-map. `values` are pulled from `meta` as required for channel inputs, `meta` is recreated / updated from channel outputs. This largely enables flow-control at the sample level, when combined with `branch`, `filter` and related operators. In some cases, `joins` cannot be avoided (or would need to be traded for concurrent processes), but I tried to minimise them. 
This means that the pipeline is essentially free of `if { }` statements, since flow control is done via channels. This also means that the pipeline DAG is always rendered in full, irrespective of whether the nodes will actually be visited by a sample. It might be possible to optimise DAG rendering by inspecting all created meta-maps and creating some global variables based on their content, to again conditionally include subworkflows, but this is beyond the scope of this refactor.

I made an effort to provide flexibility in combining assemblers, polishers, or scaffolding tools, as I thought it was reasonable, but this does not offer full-factorial combinations, which become especially tricky if things should happen in order.

Generally, I am very happy with this approach, it offers great flexibility, and is surprisingly nice to write once `meta` has been constructed. Given that global parameters can still be set, the samplesheet may or may not be kind of wide, for a single sample everything could be done via `params` and the only column in the samplesheet would be `sample`. In the most ridiculous case of setting all parameters differently for each sample, the samplesheet could grow to around 50 columns.

**Grouping**

Grouping is implemented as an extension of "`meta`-smuggling", essentially smuggling multiple meta maps through a channel (inside `meta`), replacing the value of `meta.id` with the group id.
Here is the relevant code for grouping and un-grouping while maintaining meta-maps:

```nextflow
    some_channel
         // Move group information into channel, if it exists
        .filter { it -> it.meta.group }
        .map { it -> [it.meta, it.meta.group, it.meta.ontreads] }
        // Group by group
        .groupTuple(by: 1)
        // Collect all sample-meta into a group meta slot named metas
        // Use unique reads; user responsible to group correctly
        .map {
            it ->
                [
                    [
                        id: it[1], // the group
                        metas: it[0]
                    ],
                    it[2].unique()[0] // Ontreads
                ]
        }
```
After this input channel has been processed, the samples are
    recreated from meta[metas]:

```nextflow
    process.OUT
        // Take samples with metas in slot [0]
        .filter { it -> it[0].metas }
        .flatMap { it ->
            // $it looks like [meta, output_path]
            // recreate meta from metas and update path.
            it[0].metas
                  .collect { meta -> [
                                meta: meta - meta.subMap("ontreads") + [ontreads: it[1]]
                                ]
                            }
        }
        .mix(
            process.OUT
                .filter { it -> !it[0].metas }
        )
```

*More*

Since switching to `meta`-stuffing made things much easier, I have also added an `HiC` scaffolding subworkflow, and support for `dorado polish` (experimental, as `dorado polish` does not work reliably) in the polishing subworkflow


nschan:
**This comment is outdated, and this information is now also in the PR text.**

Since the initial start of this refactor, I noticed that when doing multiple assemblies from the same set of reads it is kind of a waste to send those reads through preprocessing multiple times. I also figured that assemblies from the same set of reads are likely to be compared to each other. To reduce redundant work and make comparisons easier, there is now a `group` parameters, that can be used to put samples using the same reads into a group. Note: putting samples with different inputs into the same group will give wrong results.
I also noticed over the course of refactoring that there is some information that simply needs to be passed to processes (mostly for config purposes), since fetching those values from `params` kind of defeats the idea of parameterising per-sample. Those get stuffed into `meta` as needed. Generally, I am storing all information in the channel-map during "transit" between processes. Overall, this results in a lot of not super-concise channel manipulation. 
I have now run some more extensive tests, and overall the logic seems to work as intended.

.nf-test.log
nvnieuwk, line 1: This file should not be committed, can you remove it?

NOTE.md
nvnieuwk, line 1: Is the note meant to be in a released version?

docs/params.md
nvnieuwk, line 1: This file is not needed in nf-core pipelines, you can find all parameters on the website: https://nf-co.re/genomeassembler/dev/parameters/

docs/usage.md
```diff
@@ -11,19 +11,56 @@ This pipeline can perform assembly, polishing, scaffolding and annotation lift-o
...
+```
+
+or as **sample parameters**, by adding a _correctly named_ column to the samplesheet. In the above example this would be a column named `paramname`.
+
+Sample parameters take priority over global parameters, if both are provided the sample-specific parameter will be used for that sample.
+
+> [!NOTE]
+> The parameter names will be used in subsequent sections. Since all parameters can be provided per-sample or pipeline wide, no examples will be given.
+
+The list of all parameters that can be provided globally is available [here](params.md), parameters that can be set per sample are provided at the [end of this page](#sample-parameters).
```
nvnieuwk:
```suggestion
The list of all parameters that can be provided globally is available [here](https://nf-co.re/genomeassembler/parameters/), parameters that can be set per sample are provided at the [end of this page](#sample-parameters).
```

modules/local/collect_reads/main.nf
```diff
@@ -18,7 +18,7 @@ process COLLECT_READS {
     def prefix = task.ext.prefix ?: "${meta.id}"
 
     """
-    zcat ${reads} | gzip > ${prefix}_all_reads.fq.gz
+    cat ${reads} > ${prefix}_all_reads.fq.gz
```
nvnieuwk, line 21: This looks weird, it creates a file that ends with `.gz` but isn't a gzipped file

schema.md
nvnieuwk, line 1: I think this file was also left in by accident?

subworkflows/local/assemble/main.nf
```diff
@@ -11,211 +11,398 @@ include { QC } from '../qc/main'
...
+    // Flye:
+    ch_main_assemble_flye
+        // Convert to list for join
+        .map { it -> it.collect { entry -> [ entry.value, entry ] } }
+        .join( FLYE.out.fasta
+                .map { meta, assembly -> [meta: [id: meta.id], flye_assembly: assembly ] }
+                .map { it -> it.collect { entry -> [ entry.value, entry ] } }
+        )
+        // After joining re-create the maps from the stored map
+        .map { it -> it.collect { _entry, map -> [ (map.key): map.value ] }.collectEntries() }
```
nvnieuwk: What's the reason for recreating the maps here?

subworkflows/local/assemble/main.nf
nvnieuwk, line 1: Maybe a little note here: I found it really hard to read this code since it contained a lot of verbose parts and multiple map operator that followed eachother, it's fine for now but it will be easier to get other people to help maintain the pipeline if the code was a bit more readable

Some tips are:
- Try not to use `it` in closure but use some names that desribe the content of the channel up until that point
- Try to do everything in one `map` operator if you have multiple ones in a row. Use some nice variable names to describe each step.

nvnieuwk (commented):
Hi I've done my best but I found it pretty hard to read the code in this pipeline. so I can't approve this at this point... I've left a few comments. Here are some more tips to help with the readability:

1. Don't use `it` in closures, try to set a variable name for each item in the channel entry instead (e.g. instead of `.map { it -> ...}` do `.map { meta, file1, file2 -> ...}`. This makes it easier for me to understand what is in the channel at that point and will make it easier for future you (and others) to work on the pipeline later.
2. Use more clear variable names
3. Try to put some more comments above big code blocks with a short explanation of what this piece of code is for. (Especially on harder to understand pieces of code). 

But anyways, I'm still really impressed with what you've done here and this really will be a massive improvement to the pipeline!

NOTE.md
nschan: No, this note was mainly for me / maybe aid review.

subworkflows/local/assemble/main.nf
nschan: The reason for recreating the maps is that I heavily depend on the keys, since the "main" channel, contains around 50 elements, and I do not trust them to always be in the same slot of the list, so I use the the keys to pull them out when I need them etc.
The issue is that I need to go from map to list type for join..

subworkflows/local/assemble/main.nf
nschan: It would be nice to use names, but as I understand this would only work if I name _all_ elements of the channel / map / list in the `map` closure, and the main channel (type map) contains a lot of elements, so I am relying largely on their keys. I will try to clean up the maps / add more comments..

modules/local/collect_reads/main.nf
nschan: I think the idea here is that `${reads}` are many `f(ast)q.gz` files (via input validation), as I understand concatenated gzipped files are a valid gzip?

NOTE.md
nvnieuwk: Allright, you can choose whether to leave it in or not :)

modules/local/collect_reads/main.nf
nvnieuwk: Ah okay I haven't thought about it in that way :thinking: An interesting use case here! But I'm fine with it if it works :p

subworkflows/local/assemble/main.nf
nvnieuwk: you don't need to have a list as the key in a join operator, a map also works. I think it makes more sense to do it on the map and it will decrease the overhead code-wise a lot

subworkflows/local/assemble/main.nf
nvnieuwk: Thank you! Let me know once that has been done and I will give it another go

subworkflows/local/assemble/main.nf
nschan: It is not about the key for joining, it is that the whole channel is of map type and the join operator can only join channels of list type.

nvnieuwk: Ah okay I missunderstood your answer earlier, my bad! I'm curious if there isn't a better way to do this as this method requires a lot of overhead (having to extract the keys every time you want to join :thinking:)

nschan: Yeah, its kinda bad. I am not aware of a better way, according to @bentsherman there are plans to have `.join()` operate on maps, there is probably a way to stuff all of this into a function to hide the mess a bit?

nvnieuwk: Anyways, it's okay for me if you just add some more comments then. This can still be fixed later :)

.github/workflows/nf-test.yml
```diff
@@ -28,8 +35,8 @@ jobs:
   nf-test-changes:
     name: nf-test-changes
     runs-on: # use self-hosted runners
-      - runs-on=${{ github.run_id }}-nf-test-changes
-      - runner=4cpu-linux-x64
+      - runs-on=$-nf-test-changes
+      - runner4cpu-linux-x64
```
mashehu:
```suggestion
      - runner=4cpu-linux-x64
```

.github/actions/nf-test/action.yml
```diff
@@ -56,6 +56,8 @@ runs:
         channel-priority: strict
         conda-remove-defaults: true
 
+      # TODO Skip failing conda tests and document their failures
```
vagkaratzas: Is this TODO forgotten here by mistake? Remove if so before release, or create a GitHub issue.

.github/workflows/release-announcements.yml
```diff
@@ -34,7 +34,7 @@ jobs:
   bsky-post:
     runs-on: ubuntu-latest
     steps:
-      - uses: zentered/bluesky-post-action@6461056ea355ea43b977e149f7bf76aaa572e5e8 # v0.3.0
+      - uses: zentered/bluesky-post-action@4aa83560bb3eac05dbad1e5f221ee339118abdd2 # v0.2.0
```
vagkaratzas: Shouldn't the latest be the v0.3.0 one?

conf/modules.config
```diff
@@ -14,15 +14,14 @@ process {
...
-        mode: params.publish_dir_mode,
-        saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
+        mode: params.publish_dir_mode
     ]
 }
 
 // Read preparation
 includeConfig 'modules/ont-prep.config'
 includeConfig 'modules/hifi-prep.config'
-includeConfig 'modules/trimgalore.config'
```
vagkaratzas, line 25: I can see the reason of splitting configs, but I still think I would prefer to have all the module configurations in one file. Will let you decide, and/or the second reviewer :P

vagkaratzas (commented):
The review will be coming in separate comments, ofc :P

- What is this `assets/report` folder? Is there not a more nf-core place to put/execute those?
- Need a workflow dark, and SVGs for both light and dark metro maps

docs/output.md
```diff
@@ -61,38 +63,46 @@ If the basecalls are scattered across multiple files, `collect` can be used to c
...
+      - `<SampleName>.fastp.json`: json report
+      - `<SampleName>.fastp.log`: logfile
+    - `meryl/`: output from meryl
+      - `count/`: k-mer counts per file
+      - `unionsum/`: union of k-mer counts per sample
 
 </details>
 
-#### Short reads
+### HiC reads
```
vagkaratzas: I think this is also with  ####, seeing the identation in the menu above
```suggestion
#### HiC reads
```

docs/usage.md
```diff
@@ -7,23 +7,62 @@
...
+This pipeline can perform assembly, polishing, scaffolding using long-reads, HiC data, or a reference, and annotation lift-over from a reference genome.
 
-![Pipeline metromap](images/genomeassembler.light.png)
+> [!NOTE]
+> Phasing is currently not supported.
+
+![Pipeline metromap](images/genomeassembler_v2.light.png)
+
+Since it is often difficult to know which tool, or assembly strategy will perform best on a dataset, `nf-core/genomeassembler` can also be used to compare outcomes of different approaches in one run.
+To compare different samples, a column named `group` is required, which should contain the same value for all samples that should be compared to each other.
```
vagkaratzas: Two unknown characters at the end of this line.

.github/workflows/release-announcements.yml
nschan: It probably should, yes

.github/actions/nf-test/action.yml
nschan: Forgotten, will remove it

README.md
```diff
@@ -10,7 +10,7 @@
...
 [![nf-test](https://img.shields.io/badge/unit_tests-nf--test-337ab7.svg)](https://www.nf-test.com)
 
-[![Nextflow](https://img.shields.io/badge/version-%E2%89%A525.04.0-green?style=flat&logo=nextflow&logoColor=white&color=%230DC09D&link=https%3A%2F%2Fnextflow.io)](https://www.nextflow.io/)
+[![Nextflow](https://img.shields.io/badge/version-%E2%89%A525.10.0-green?style=flat&logo=nextflow&logoColor=white&color=%230DC09D&link=https%3A%2F%2Fnextflow.io)](https://www.nextflow.io/)
 [![nf-core template version](https://img.shields.io/badge/nf--core_template-3.5.1-green?style=flat&logo=nfcore&logoColor=white&color=%2324B064&link=https%3A%2F%2Fnf-co.re)](https://github.com/nf-core/tools/releases/tag/3.5.1)
 [![run with conda](http://img.shields.io/badge/run%20with-conda-3EB049?labelColor=000000&logo=anaconda)](https://docs.conda.io/en/latest/)
 [![run with docker](https://img.shields.io/badge/run%20with-docker-0db7ed?labelColor=000000&logo=docker)](https://www.docker.com/)
```
vagkaratzas, line 16: First thing in the `## Introduction` section, the workflow map should be shown! You went into the trouble of making that look so nice, make sure people see it :D
Definitely better here than in the `usage.md` you currently have it!! It is the first thing that the users will see on the website and on the GitHub page.
Include something like this under the `Introduction` section:
```
<p>
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="docs/images/proteinfamilies_workflow_dark.png">
    <img alt="nf-core/proteinfamilies workflow overview" src="docs/images/proteinfamilies_workflow_light.png">
  </picture>
</p>
```
( where instead of proteinfamilies you have your links, ofc :P )

conf/modules.config
nschan: I can see why you would prefer this, but in practice, having a massive module configuration file turned out to be very hard for me manage and I spent a lot of time looking for specific configurations in there and kept introducing duplications etc. There are around 70 `withName` selectors across the config files.

NOTE.md
```diff
@@ -0,0 +1,51 @@
+branch refactor-assemblers:
```
vagkaratzas: Yep, remember to remove this. And/or move stuff to `usage.md` if needed.

nf-test.config
```diff
@@ -1,4 +1,4 @@
-config {
+    config {
```
vagkaratzas:
```suggestion
config {
```

nschan:
> The review will be coming in separate comments, ofc :P

ok ;)

> * What is this `assets/report` folder? Is there not a more nf-core place to put/execute those?

assets/report contains the scripts for generating the report. If there is a better place to put them I am happy to put them somewhere else, but idk where. 

> * Need a workflow dark, and SVGs for both light and dark metro maps

Good point, I will add those.

CHANGELOG.md
```diff
@@ -3,6 +3,53 @@
...
+- HiC scaffolding subworkflow:
+  - mapping with `bwamem2` or `minimap2`
+  - duplicate removal with `picard`
+  - scaffolding with `yahs`
+- Switched to the versions topic, requires nextflow >=25.10.0
+
+### `Fixed`
+
+### `Dependencies`
+
```
vagkaratzas, line 59: Can you use the nf-core dependencies table for this, along with versions:
example:
```
| Tool    | Previous version | New version |
| ------- | ---------------- | ----------- |
| multiqc | 1.32             | 1.33        |
| cmaple  | -                | 1.1.0       |
```

CHANGELOG.md
```diff
@@ -3,6 +3,53 @@
...
+This is a major release, with breaking changes.
+v2.0.0 of genomeassembler is a large refactor of the pipeline to facilitate sample-level parameteristation. This allows to either parameterise the _pipeline_ using `params`, or parameterise _samples_ via the `input` samplesheet. In case both types of parameterisations are used, sample parameters will take priority.
+
+Since this workflow follows a sample-centric implementation, nextflow will always render the full pipeline dag, but depending on configuration samples may not travel through the whole pipeline. This may also cause terminal output to show task instances that will never become an active process.
+
+In addition, v2.0.0 contains these changes:
+
+### `Added`
+
+- fastplong for long-read trimming and qc
```
vagkaratzas, line 24: I know it is a bit difficult to do now, but in general it would be nice to include the PR# along each added feature (and the git handle of the person(s) who worked on it)

CHANGELOG.md
```diff
@@ -3,6 +3,53 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## v2.0.0 'Saffron Vulture' - [2026-xx-xx]
```
vagkaratzas:
```suggestion
## v2.0.0 - 'Saffron Vulture' - [2026-xx-xx]
```

workflows/genomeassembler.nf
```diff
@@ -195,54 +216,36 @@ workflow GENOMEASSEMBLER {
...
-            sort: true,
-            newLine: true
-        )
+    channel
+        .fromPath("${projectDir}/assets/report/scripts/*")
+        .collect()
+        .set { report_scripts }
+
+    //fastplong_jsons.view { it -> "UNQIE JSONS: $it"}
+
```
vagkaratzas:
```suggestion
```

workflows/genomeassembler.nf
```diff
@@ -195,54 +216,36 @@ workflow GENOMEASSEMBLER {
...
+    REPORT( report_files,
```
vagkaratzas, line 207: Nothing for multiqc? :O

workflows/genomeassembler.nf
```diff
@@ -3,153 +3,168 @@
...
-    ASSEMBLE.out.longreads.set { ch_longreads }
-    ch_versions = ch_versions.mix(ASSEMBLE.out.versions)
     /*
     Polishing
     */
+    ch_main_assembled
+        .branch {
+            it ->
+            def polishers = ["pilon", "medaka", "medaka+pilon", "dorado", "dorado+pilon"]
+            /*debug
```
vagkaratzas: remove debug?

workflows/genomeassembler.nf
```diff
@@ -3,153 +3,168 @@
...
-    */
-    if (params.hifi) {
-        HIFI(ch_input)
-        HIFI.out.hifi_reads.set { ch_hifi_reads }
+    PREPARE.out.meryl_kmers.set { meryl_kmers }
 
-        ch_versions = ch_versions.mix(HIFI.out.versions)
-    }
 
+    //ch_main_prepared.view{"Main WF: Prepared out: $it "}
```
vagkaratzas: debug leftovers?

vagkaratzas (commented):
Another round. Hopefully another last one soon!

workflows/genomeassembler.nf
nschan: I am not using multiqc for reporting

CHANGELOG.md
nschan: How do I get this nf-core dependencies table?

CHANGELOG.md
nschan: Yes, fully agree. This all a massive PR, sadly there is only one person working on this (me).

modules/local/dorado/aligner/main.nf
```diff
@@ -0,0 +1,43 @@
...
+
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+
+    """
+    dorado aligner \\
```
vagkaratzas, line 24: Is this hard to push to nf-core?

README.md
nschan: Isn't that what is in line 27-30 of this file? I need to update the figure though to have a dark version.

modules/local/genomescope/main.nf
```diff
@@ -14,7 +14,8 @@ process GENOMESCOPE {
         tuple val(meta), path("*_plot.log.png")     , emit: plot_log
```
vagkaratzas, line 14: `https://nf-co.re/modules/genomescope2/` 
is on nf-core/modules, update and use that one, or maybe just patch it?

modules/local/gfa2fa/main.nf
```diff
@@ -2,37 +2,27 @@ process GFA_2_FA {
     tag "${meta.id}"
```
vagkaratzas: [gfatools_gfa2fa](https://nf-co.re/modules/gfatools_gfa2fa/) also on nf-core modules

modules/local/jellyfish/count/main.nf
```diff
@@ -10,7 +10,8 @@ process COUNT {
```
vagkaratzas: https://nf-co.re/modules/jellyfish_count/ also (I made that one >.<)

modules/local/jellyfish/dump/main.nf
```diff
@@ -10,24 +10,16 @@ process DUMP {
 
     output:
```
vagkaratzas, line 11: And `dump` https://nf-co.re/modules/jellyfish_dump/ xD

modules/local/jellyfish/histo/main.nf
```diff
@@ -10,25 +10,17 @@ process HISTO {
```
vagkaratzas, line 10: If you update the others, you can add this one too :P

modules/local/longstitch/main.nf
```diff
@@ -12,7 +12,8 @@ process LONGSTITCH {
     output:
```
vagkaratzas: https://nf-co.re/modules/longstitch/

modules/local/medaka/medaka_consensus/main.nf
```diff
@@ -13,7 +13,7 @@ process MEDAKA_PARALLEL {
```
vagkaratzas, line 13: I also see the medaka one: https://github.com/nf-core/modules/blob/master/modules/nf-core/medaka/main.nf

modules/local/quast/main.nf
```diff
@@ -15,7 +15,7 @@ process QUAST {
     output:
```
vagkaratzas: quast also in nf-core. If you ened minor changes from the nf-core ones, you usally install them from nf-core, make the changes, and then use the `nf-core modules patch` command to keep the diff file.

modules/local/report/main.nf
```diff
@@ -10,73 +10,73 @@ process REPORT {
     https://wave.seqera.io/view/builds/bd-be4a8863b7b76cf7_1 docker
```
vagkaratzas, line 10: I guess this is your local alternative to MultiQC?

subworkflows/local/utils_nfcore_genomeassembler_pipeline/main.nf
```diff
@@ -95,50 +95,220 @@ workflow PIPELINE_INITIALISATION {
...
+        .set { ch_samplesheet }
+
+    // Define valid hybrid assemblers
+
+    def hybrid_assemblers = ["hifiasm"]
+    ch_samplesheet.dump(tag: "PARSED INPUTS:")
+    // sample-level checks
+    // if a check fails, map returns a list that prints what fails, and contains "invalid"
+    // error is raised by subscribe if there is more than one "invalid"
+    /*
```
vagkaratzas: Keep or remove this comment block?

subworkflows/local/scaffold/main.nf
```diff
@@ -1,76 +1,210 @@
-include { RUN_LINKS } from './links/main'
-include { RUN_LONGSTITCH } from './longstitch/main'
-include { RUN_RAGTAG } from './ragtag/main'
+include { RUN_LINKS         } from './links/main'
+include { RUN_LONGSTITCH    } from './longstitch/main'
+include { RUN_RAGTAG        } from './ragtag/main'
+include { HIC               } from './hic/main'
 
 workflow SCAFFOLD {
```
vagkaratzas, line 6: I would keep the workflow name same to the folder name (either choose `scaffolding` or `scafold` for both.

subworkflows/local/scaffold/main.nf
```diff
@@ -1,76 +1,210 @@
...
-
-    if (params.scaffold_ragtag) {
-        RUN_RAGTAG(inputs, in_reads, assembly, references, ch_aln_to_ref, meryl_kmers)
-        RUN_RAGTAG.out.busco_out.set { ragtag_busco }
-        RUN_RAGTAG.out.quast_out.set { ragtag_quast }
-        RUN_RAGTAG.out.merqury_report_files.set { ragtag_merqury }
-
-        ch_versions = ch_versions.mix(RUN_RAGTAG.out.versions)
-    }
+    channel.empty().set { links_busco }
```
vagkaratzas: I think the preferred nf-core way is 
```suggestion
    ch_links_busco = channel.empty()
```
instead of `.set`. But I guess too late for that now, given the number of files in teh PR xD

subworkflows/local/scaffold/longstitch/main.nf
```diff
@@ -1,45 +1,68 @@
...
+                ]
+        }
+        .set { liftoff_in }
 
-    versions = ch_versions
+    LIFTOFF(liftoff_in,[])
 
     emit:
-    scaffolds
+    ch_main                 = ch_main_scaffolded
```
vagkaratzas, line 60: `main` is not very self-explanatory as a name

subworkflows/local/scaffolding/longstitch/main.nf
```diff
@@ -1,45 +1,68 @@
...
-    in_reads
-    assembly
-    _references
-    ch_aln_to_ref
+    ch_main
     meryl_kmers
-    genome_size
+
+    /*
+    TODO:
```
vagkaratzas: remove TODO?

subworkflows/local/scaffold/hic/main.nf
```diff
@@ -0,0 +1,142 @@
+include { QC                                        } from '../../qc/main'
+include { YAHS                                      } from '../../../../modules/nf-core/yahs/main'
+include { RUN_LIFTOFF                               } from '../../liftoff/main'
+include { BWAMEM2_MEM                               } from '../../../../modules/nf-core/bwamem2/mem/main'
+include { BWAMEM2_INDEX                             } from '../../../../modules/nf-core/bwamem2/index/main'
+include { SAMTOOLS_FAIDX                            } from '../../../../modules/nf-core/samtools/faidx/main'
+include { MINIMAP2_ALIGN as MINIMAP2_HIC            } from '../../../../modules/nf-core/minimap2/align/main'
+include { PICARD_MARKDUPLICATES as MARKDUP          } from '../../../../modules/nf-core/picard/markduplicates/main'
+include { PICARD_ADDORREPLACEREADGROUPS as ADD_RG   } from '../../../../modules/nf-core/picard/addorreplacereadgroups/main'
+workflow HIC {
```
vagkaratzas, line 12:
```suggestion

workflow HIC {
```

subworkflows/local/prepare/main.nf
```diff
@@ -0,0 +1,256 @@
...
+    SHORTREADS, JELLYFISH, ONT and HIFI each implement
+    the same logic for sample grouping.
+    Grouping needs to be specified by the user, and can
+    be used to create sample groups that share inputs, to
+    minimize redundant input preparations.
+    Reads of samples from the same group will be prepared
+    only once, and then the original channel is restored.
+
+    Brief description how this works:
+        // Move group information into channel, if it exists
```
vagkaratzas, line 18: Too many comments and code there. I guess it should be deleted? The explanation should be enough I guess

vagkaratzas (changes requested):
Final one. There is no way that everything will be working as intended with all these changes (and strategies..!)
I would suggest creating nf-test for all local subworkflows and modules if they dont already have. But I wouldn't stop you from a release for that.
BUT! I think this is a great opportunity to have a track for the upcoming hackathon for beginners and not only, to write nf-tests for everything. Just a suggestion ;)

subworkflows/local/scaffold/longstitch/main.nf
nschan: Fair point. I am not sure how to better name this, since the `main` channel does not necessarily contain all steps for all samples. I think I am using `main` to mean that this has been transformed correctly and that everything is ready to move back into the main workflow / next subworkflow.

subworkflows/local/scaffold/main.nf
nschan: If it is preferred, but not mandatory, I would like to keep it with `.set`. I know some people prefer `channel =` but its ugly :/

subworkflows/local/prepare/main.nf
nschan: This comment is mainly for reviewers and potential future contributors. If this is useless, I can remove it.

modules/local/jellyfish/count/main.nf
nschan: A, great, I wasn't aware that you made an nf-core module, happy to switch!

modules/local/longstitch/main.nf
nschan: Guess I could have been aware of that one..

modules/local/report/main.nf
nschan: Exactly

modules/local/dorado/aligner/main.nf
nschan: Probably not, I stole this from https://github.com/nf-core/methylong/tree/2.0.0/modules/local/dorado

CHANGELOG.md
vagkaratzas: Manual labor xD

README.md
vagkaratzas: Oh it's there already! Missed that!

nschan:
> Final one. There is no way that everything will be working as intended with all these changes (and strategies..!) I would suggest creating nf-test for all local subworkflows and modules if they dont already have. But I wouldn't stop you from a release for that. BUT! I think this is a great opportunity to have a track for the upcoming hackathon for beginners and not only, to write nf-tests for everything. Just a suggestion ;)

I would hope that everything works as intended and I did some larger tests before asking for reviews. Still it is likely that someone will try something that I did not think of; `full_test` tests a bunch of strategies. Having a track for this at the hackathon would be nice, but for me this overlaps with teaching obligations and I will not be able to participate myself. I am not sure how productive putting up a bunch of stuff without being involved would be?

subworkflows/local/utils_nfcore_genomeassembler_pipeline/main.nf
nschan: I will look into this block again, to see if it should be kept. I currently favour the approach that happens on l145; validating before constructing meta. This block is a bit old..

vagkaratzas:
> > Final one. There is no way that everything will be working as intended with all these changes (and strategies..!) I would suggest creating nf-test for all local subworkflows and modules if they dont already have. But I wouldn't stop you from a release for that. BUT! I think this is a great opportunity to have a track for the upcoming hackathon for beginners and not only, to write nf-tests for everything. Just a suggestion ;)
> 
> I would hope that everything works as intended and I did some larger tests before asking for reviews. Still it is likely that someone will try something that I did not think of; `full_test` tests a bunch of strategies. Having a track for this at the hackathon would be nice, but for me this overlaps with teaching obligations and I will not be able to participate myself. I am not sure how productive putting up a bunch of stuff without being involved would be?

I would ask the thematically closest persons / facility / nf-core slack channels for potential track leaders that would want to oversee this if they don't already have a project but feel confident enough with nf-tests.

subworkflows/local/scaffold/main.nf
prototaxites: Concur that `.set { }` is hard to parse and follow...! I would recommend switching (`.set` is going to be removed from the operator list in an upcoming version of Nextflow in any case)

OlivierCoen: Actually, the Nextflow best-practice is now to use `channel =` instead of `set`:
https://www.nextflow.io/docs/latest/reference/operator.html#set

nschan: Okay, well, if using `set` breaks type checking I guess it makes sense to switch to `=`.

CHANGELOG.md
d4straub: At least each PR should be referenced here, I agree.

CHANGELOG.md
d4straub: Proper reporting includes version changes imho, so I see that as mandatory.

assets/report/functions/read_genomescope.R
```diff
@@ -1,43 +1,44 @@
+## WIP
```
d4straub: Still WIP?

assets/report/scripts/_busco.R
d4straub, line 1: These scripts seem rather a lot of investment, including maintenance. Potentially it would be easier to use for some (standard) programs like busco rather MultiQC? Just a thought, not needed to change now.

Generally it seems to me that it could be easier for other interested parties or even future you to include more comments intothat scripts. I thought also about input/output, but maybe at least for what tool version the script was written, so that its possible to track down expected inputs easier.

conf/modules/assembly.config
d4straub, line 1: I am not a huge fan of these split configs, but I also dont really find an important disadvantage.
But I am a little concerned that I cannot really see a clear pattern. Sometimes its a single module, sometimes several. If it would be grouped by something intuitive, for example subworkflow or such, then I would find it more appealing.

conf/modules/assembly.config
```diff
@@ -3,57 +3,56 @@ process {
         ext.args = {
             [
                 meta.genome_size ? "--genome-size ${meta.genome_size}" : '',
-                params.flye_args
+                // If flye is both for ONT and HiFi, we need to detect the type of read based on the read suffix (fastplong)
+                // This is presumably prone to fail in certain weird situations
+                (meta.assembler_ont == "flye" && meta.assembler_hifi == "flye" && strategy == "scaffold") ?
+                (reads ==~ ".*${meta.id}_ont.*" ? meta.assembler_ont_args : meta.assembler_hifi_args) :
+                (meta.assembler_ont == "flye") ? meta.assembler_ont_args : "",
```
d4straub: Argh. Maybe at least indentation?

```suggestion
                (meta.assembler_ont == "flye" && meta.assembler_hifi == "flye" && strategy == "scaffold") ?
                    (reads ==~ ".*${meta.id}_ont.*" ? meta.assembler_ont_args : meta.assembler_hifi_args) :
                        (meta.assembler_ont == "flye") ? meta.assembler_ont_args : "",
```
This seems not very transparent though :D

conf/modules/assembly.config
```diff
@@ -3,57 +3,56 @@ process {
         ext.args = {
             [
                 meta.genome_size ? "--genome-size ${meta.genome_size}" : '',
-                params.flye_args
+                // If flye is both for ONT and HiFi, we need to detect the type of read based on the read suffix (fastplong)
+                // This is presumably prone to fail in certain weird situations
```
d4straub: Why? What part? In what situation? What would be the consequence of this failing?
If you can anticipate a problem, it might be worthwhile to describe with a little detail so that its easier to come back and solve it in case it indeed leads to trouble.

conf/modules.config
d4straub: If it serves a purpose and the split follows a transparent pattern, then I am fine with it.

conf/test.config
```diff
@@ -21,13 +21,11 @@ process {
...
-    jellyfish = false
-    genome_size = 2000000
-    hifi = true
-    ont = true
-    assembler = "flye_on_hifiasm"
+    input = 'https://raw.githubusercontent.com/nf-core/test-datasets/genomeassembler/samplesheet/test_samplesheet_v2.csv'
+    ontreads = 'https://raw.githubusercontent.com/nf-core/test-datasets/genomeassembler/A_thaliana_Col-0_2mb/ONT-Col-0_test_data.fastq.gz'
+    hifireads = 'https://raw.githubusercontent.com/nf-core/test-datasets/genomeassembler/A_thaliana_Col-0_2mb/HiFi-Col-0_test_data.fastq.gz'
+    ref_fasta = 'https://raw.githubusercontent.com/nf-core/test-datasets/genomeassembler/A_thaliana_Col-0_2mb/Col-CEN_v1.2.Chr1_5MB-7MB.fasta.gz'
+    ref_gff = 'https://raw.githubusercontent.com/nf-core/test-datasets/genomeassembler/A_thaliana_Col-0_2mb/Col-CEN_v1.2_genes_araport11.Chr1_5MB-7MB.gff3.gz'
```
d4straub: Afaik we are using here typically `params.pipelines_testdata_base_path` (not sure whats the gain though), do you have a reason not to do that?

https://github.com/nf-core/ampliseq/blob/b4d98050ada6c133dbab0f1983de90b0bc9c670b/conf/test.config#L28-L29

docs/images/genomeassembler_v2.light.png
d4straub, line 1: I am sensitive to bad alignment, and I am seeing here non-centered header for each box, boxes not aligned (or the box header, not sure). Otherwise appears pretty to me!

docs/usage.md
```diff
@@ -7,23 +7,62 @@
...
+
+> [!WARNING]
+> Grouping should **never** be used for samples that use different input files.
+
+## Choice of assembly-strategy and assembler
 
-### Pre-set profiles
+Assembly strategy is controlled via `strategy` (either pipeline parameter or sample-setting), and assembler(s) used are chosen via `assembler`
 (either pipeline parameter or sample-setting)
+`nf-core/genomeassembler` currently supports the following assembly strategies:
```
d4straub: There is something weird going on on github at least: 

Image

docs/usage.md
```diff
@@ -55,13 +95,11 @@ Further columns _can_ be:
   - `shortread_R`: shortread reverse file (paired end)
   - `paired`: [true/false] true if the reads are paired end, false if they are single-end. The `shortreads_R` column should exist if `paired` is `false` but can be empty.
 
+But samplesheets can grow more complex if a range of strategies should be compared in a single pipeline run. A list of all possible columns can be found at the [end of this page](#sample-parameters)
+
 > [!INFO]
 > It is strongly recommended to provide all paths as absolute paths
```
d4straub, line 101: Deosnt render, not sure why not

docs/usage.md
```diff
@@ -221,3 +259,259 @@ We recommend adding the following line to your environment to limit this (typica
...
+| ----------- | ------------------------------------------ | --------- |
+| `ref_fasta` | Path to reference genome seqeunce (fasta)  | `string`  |
+| `ref_gff`   | Path to reference genome annotations (gff) | `string`  |
+| `use_ref`   | Use reference genome                       | `boolean` |
+
+## Assembly options
+
+Options controlling assembly.
+
+> NOTE: hifiasm_args and flye_args will be passed to the respective assembler in all cases. If the same assembler is used in different strategies, these need may need to be parameterised per sample.
```
d4straub: different syntax then other notes, I would unify

modules/local/dorado/polish/main.nf
```diff
@@ -0,0 +1,42 @@
+process DORADO_POLISH {
+    tag "${meta.id}"
+    label 'process_high'
+
+    container "docker.io/nanoporetech/dorado:sha00aa724a69ddc5f47d82bd413039f912fdaf4e77"
```
d4straub, line 5: not even conda? So this part doesnt work with conda at all, or even the whole pipeline?

modules/local/dorado/polish/main.nf
```diff
@@ -0,0 +1,42 @@
...
+    input:
+    tuple val(meta), path(assembly), path(alignment), path(index)
+    val(variant_call_format)
+
+    output:
+    tuple val(meta), path("${meta.id}_dorado_polished.fa.gz"), emit: polished_alignment, optional: true
+    tuple val(meta), path("${meta.id}_dorado_polished*vcf"), emit: variant_calls, optional: true
+    tuple val("${task.process}"), val('dorado'), eval('dorado --version 2>&1 | head -n1'), emit: versions_dorado, topic: versions
+
+
```
d4straub: I dislike unnecessarily empty lines 🗡️ 

```suggestion
```

modules/local/dorado/polish/main.nf
```diff
@@ -0,0 +1,42 @@
...
+        ${args} \\
+        ${variants} \\
+        ${outfile}
+    """
+
+    stub:
+    def args   = task.ext.args ?: ''
+    def variants = ["vcf","gvcf"].contains(variant_call_format) ? "--${variant_call_format}" : ""
+    def outfile = variants ? "touch ${meta.id}_dorado_polished.${variants}" : "echo '' | bgzip > ${meta.id}_dorado_polished.fa.gz"
+
```
d4straub:
```suggestion
```

subworkflows/local/prepare/main.nf
d4straub: I would keep it. nothing wrong with detailed explanation. I wouldn't know where else to put it. Well, maybe a docu directly for contributors, but I think the PR is already far from small.

subworkflows/local/utils_nfcore_genomeassembler_pipeline/main.nf
```diff
@@ -95,50 +95,218 @@ workflow PIPELINE_INITIALISATION {
...
+                                    null
+
+            def hic_F           =   it.hic_F ?: params.hic_F
+
+            def scaffold_hic    =   hic_F ? (it.scaffold_hic != null ? it.scaffold_hic : params.scaffold_hic) : false
+
+            def hic_trim        =   !scaffold_hic ? false :
+                                    (it.hic_trim ?: params.hic_trim)
+
+            def assembler_ont_args =  it.assembler_ont_args ?: params.assembler_ont_args ?: ''
```
d4straub, line 137: for my taste too many empty lines, generally, but especially here.

d4straub (commented):
Generally, that meta-guided strategy seems really interesting to me. Its not very transparent though. But it seems overall good..

A problem for me is that this PR is so extensive that its close to impossible to review. In my opinion, it should be merged asap. The biggest blockers to me are:
- use existing nf-core modules were possible
- split configs in pieces that follow a pattern (I didnt get the pattern)
- make CI tests pass
At this point I would like to trust in your local test runs (I havent run anything).

And later, in another PR (please **not** this one) add as many tests as you can, that will make the pipeline stable and easier to contribute and work with.

modules/local/dorado/polish/main.nf
nschan: ONT does not seem to provide dorado via conda. This would only affect this specific step, i.e. polishing with dorado. The other parts of the pipeline are not affected by this.

assets/report/scripts/_busco.R
nschan: That is a good point. I have received somewhat conflicting reviews regarding the extend of comments, so I have not commented extensively, but I would agree that more comments probably do not hurt.

conf/test.config
nschan: There was no real reason, I have changed it. I think the gain is that you can easily point e.g. to your own local test-data and do not have to fetch from github every time?

modules/local/dorado/aligner/main.nf
nschan: The main issue may be the unavailability of dorado via conda?

subworkflows/local/prepare/main.nf
nschan: I will leave this here for now and prepare an explanation / docs for contributors after this whole thing is finally merged..

conf/modules/assembly.config
nschan: I added an example. This is one of those things that feels like it would be easy to construct a case that breaks it, but that would probably require some intention, or gross misunderstanding..

CHANGELOG.md
```diff
@@ -3,6 +3,117 @@
...
+| trimgalore | 0.6.10 | deprecated               |
+| cutadapt   | 4.9    | deprecated               |
+
+The following param is no longer implemented:
+
+- `dump`, used to dump jellyfish output.
+
+#### Unchanged
+
+The following tools are unchanged:
```
d4straub, line 103: that is not really needed I think, only what changed is appearing in a changelog, imho. But fine for now as is I think.

nextflow.config
```diff
@@ -42,55 +42,81 @@ params {
...
+    hic_R                      = []
+    hic_trim                   = true
+    // == ASSEMBLY ==
+    assembler                  = "hifiasm" // assembler to use
+    assembly_scaffolding_order = "ont_on_hifi"
     // -- Assembly: Flye --
     genome_size                = null // genomesize, optional, can be estimated from ONT reads
-    flye_mode                  = '--nano-hq' // flye mode
-    flye_args                  = "" // extra flye args
+    // DEPRECATED: flye_mode                  = '--nano-hq'
```
d4straub: this desnt need to stay here I think

CHANGELOG.md
nschan: Yeah, I mainly created that so I do not have to look things up in the future ;)

nschan:
@vagkaratzas Your review is requesting changes. I am lost which ones, please let me know. I would prefer to postpone contribution of modules to nf-core and re-adding them here to a later point. This PR is almost 1 year old by now and I would like to merge it to dev and then sort other things in a more focused way. 
The nf-tests are currently running, they may have issues with snapshots, since longstitch seems to produce large non-deterministic output, if those occur I will ignore more files, or stop testing longstitch..

vagkaratzas (approved):
Substantial amount of work! Good enough for me!

d4straub (approved):
LGTM!

nschan:
Additional local testing using the full_test profile, which is a lot more extensive than the CI test, has uncovered some problems.

The test is still running, but the recent commits address issues that occured largely as a consequence of the migration of `jellyfish` and `genomescope` to nf-core modules:

  - nf-core module `jellyfish/count` does not accept gzipped inputs -> patched module in this pipeline, might be worth patching upstream? @vagkaratzas what do you think?
  - local swf `prepare/jellyfish` was adapted to include size, but this change has not been added to the group preparation. This has been added. 
  - nf-core module `genomescope2` fails to converge for the test dataset I use, but the local `genomescope` module converges fine for the identical dataset. I have re-added in the local genomescope module, as well as `bin/genomescope.R`. I have not investigated what is causing this.

vagkaratzas:
> * nf-core module `jellyfish/count` does not accept gzipped inputs -> patched module in this pipeline, might be worth patching upstream? @vagkaratzas what do you think?

Definitely, 100/100!! Tag me in the PR

> * nf-core module `genomescope2` fails to converge for the test dataset I use, but the local `genomescope` module converges fine for the identical dataset. I have re-added in the local genomescope module, as well as `bin/genomescope.R`. I have not investigated what is causing this.

Should be OK for now, but interesting to investigate more when you have the time. Maybe just create a new Issue in this repo for this task.

nschan:
The continued testing has also uncovered a very strange problem that seems to be a consequence of updating `bam_stats_samtools`, or maybe nextflow version and strictness.

In any case, this no longer worked as it used to, and now appears to require a faidx input. Generating this index requires some upstream steps (bgzip vs gzip largely).

nschan:
The fallout from the `bgzip` requirement was larger than I initially anticipated. Many modules, such as `flye` or `gfatools` make use of `gzip` and do not ship `bgzip` in the containers. For now, I have opted for explicit `bgzip` processes, and have patched the modules to not pipe into `gzip`. Based on some [discussion on slack](https://nfcore.slack.com/archives/CJRH30T6V/p1782198171532779) I think it would be good to instead patch the relevant nf-core modules to use `bgzip` mid-term.

PR #192, opened 2026-06-26
Add `BGZIP` to longstitch, remove `UNZIP` from `MAP_TO_ASSEMBLY`

### LINKED ISSUE #190: Migrate modules to `bgzip`?
The fallout from the bgzip requirement in #171 was larger than I initially anticipated. Many modules, such as flye or gfatools make use of gzip and do not ship bgzip in the containers. For now, I have opted for explicit bgzip processes, and have patched the modules to not pipe into gzip. Based on some [discussion on slack](https://nfcore.slack.com/archives/CJRH30T6V/p1782198171532779) I think it would be good to instead patch the relevant nf-core modules to use bgzip mid-term.

nschan:
This addresses some leftover from #171, and now uses bgzip explicitly to zip each output fasta. This allows indexing with `SAMTOOLS_INDEX` without decompression.

This PR closes #190


jfy133 (approved):
Just missing CHANGELOG

PR #196, opened 2026-06-29
Medaka bgzip

nschan:
Adds bgzipping of gzipped medaka outputs. Was missed in #193

### DIFF
--- a/subworkflows/local/polishing/medaka/polish_medaka/main.nf
+++ b/subworkflows/local/polishing/medaka/polish_medaka/main.nf
@@ -1,6 +1,8 @@
 include { MEDAKA_PARALLEL as MEDAKA } from '../../../../../modules/local/medaka/medaka_consensus/main'
 include { QC } from '../../../qc/main.nf'
 include { LIFTOFF } from '../../../../../modules/nf-core/liftoff/main'
+include { HTSLIB_BGZIPTABIX as BGZIP } from '../../../../../modules/nf-core/htslib/bgziptabix/main'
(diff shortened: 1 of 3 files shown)


subworkflows/local/polishing/medaka/polish_medaka/main.nf
```diff
@@ -1,6 +1,8 @@
 include { MEDAKA_PARALLEL as MEDAKA } from '../../../../../modules/local/medaka/medaka_consensus/main'
 include { QC } from '../../../qc/main.nf'
 include { LIFTOFF } from '../../../../../modules/nf-core/liftoff/main'
+include { HTSLIB_BGZIPTABIX as BGZIP } from '../../../../../modules/nf-core/htslib/bgziptabix/main'
+include { HTSLIB_BGZIPTABIX as UNZIP } from '../../../../../modules/nf-core/htslib/bgziptabix/main'
```
mashehu, line 5: why not just use the gunzip module here, than you don't need to make a `ch_to_unzip` channel but go straight `UNZIP(MEDAKA.out.assembly)` I think

mashehu (approved):
Just a small optimization (not tested on my side)

subworkflows/local/polishing/medaka/polish_medaka/main.nf
nschan: I had UNZIP in here initially, but found it a bit confusing that zip and unzip had different input formats, and figured that it would make more sense to use the same module for all zip operations, since I would anyway need `HTSLIB_BGZIPTABIX`

mashehu: makes sense

PR #197, opened 2026-06-29
Used trimmed reads with jellyfish

nschan:
Jellyfish should receive trimmed reads, and not take the raw reads. 
To do this, `qc_reads_path` needs to be updated to the trimmed reads after preparing the input files. Previously, this update happened after Jellyfish, which was too late.


mashehu (approved):
maybe add a code comment like you had before?

Issue #191, opened 2026-06-26
Fix `report` on `dev`


nschan:
After #171, the report is broken. This will be fixed after #171 is merged.

Issue #112, opened 2025-02-27
More elegant handling of versions for report


nschan:
Currently the versions for reporting are generated via `softwareVersionsToYAML`, the yaml is parsed in the report. After the report completed, `softwareVersionsToYAML` runs again to produce a yaml that also includes the tools used for the report. This is not very elegant, and there is probably a better way to get the software versions into the report.

PR #202, opened 2026-07-01
Fix report

nschan:
As described in #191, reporting broke during pipeline refactoring. 
This PR restores report functionality, by fixing some pipeline logic. This also addresses #112, which is much easier now that the pipeline is using the versions `topic`.

The `quarto` reporting done is based around a lot of conditional executions, which may be a bit hard to follow. I am providing a report from a small run (Assembly of A. thaliana Col-0 using flye) here: 
https://syncandshare.lrz.de/getlink/fiDYBkvhV8Cn7TCuXtLehX/report_ga.zip
this run was using the samplesheet below, which is subset of the pipelines `full_test` profile:

```csv
sample,strategy,assembler,assembler_ont,assembler_hifi,scaffold_ragtag,scaffold_longstitch,scaffold_links,use_ref,group,qc_reads,assembly_scaffolding_order,polish,hic_F,hic_R,hic_trim
test_ont_flye,single,flye,flye,,,,true,false,flye,,,,,,
test_hifi_flye,single,,,flye,true,,true,true,hifi_flye,hifi,,,,,
test_flye_flye,scaffold,flye_flye,,,true,,,,flye,,,pilon,,,
``` 

paths to reads were provided via params, in addition to these:

```
params {
    quast               = true
    busco               = true
    jellyfish           = true
    use_ref             = true
    shortread_trim      = true
    busco_lineage       = "brassicales_odb12"
    outdir              = '/stuff/results_flye_test'
}
```


assets/report/scripts/_quast.R
```diff
@@ -2,7 +2,7 @@
 
 # Parse the quast reports from data/quast
 quast_stats <- list.files(paste0(data_base, "quast"),
-                          pattern = "report.tsv",
+                          pattern = ".tsv",
```
mashehu:
```suggestion
                          pattern = "\\.tsv$",
```
should be saver

workflows/genomeassembler.nf
```diff
@@ -187,9 +191,11 @@ workflow GENOMEASSEMBLER {
                 SCAFFOLD.out.scaffold_merqury_reports
             )
         )
-        .collect { it -> [it[1], it[2], it[3], it[4]] }
+        .filter { it -> it != null }
+        .collect { reports -> [reports[1], reports[2], reports[3], reports[4]] }
         .toSet()
         .flatten()
+        .filter { it -> it != null }
```
dialvarezs, line 202: Is this `.filter` required? The previous one should have catched null already.

modules/local/report/main.nf
```diff
@@ -3,11 +3,11 @@ process REPORT {
     label 'process_low'
     conda "${moduleDir}/environment.yml"
     container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
-        ? 'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/89/8967e1cb830fdc77ec5b84541a50c74a0a05eaaae557314490809de2fc91e4af/data'
-        : 'community.wave.seqera.io/library/quarto_r-gt_r-plotly_r-quarto_pruned:be4a8863b7b76cf7'}"
-    /* wave builds:
-    https://wave.seqera.io/view/builds/bd-6e20dd9b9b77f359_1 singularity
-    https://wave.seqera.io/view/builds/bd-be4a8863b7b76cf7_1 docker
+        ? 'oras://community.wave.seqera.io/library/quarto_r-gt_r-plotly_r-quarto_pruned:c7c55c1c911608e9'
```
mashehu: please use https here.

if you get the latest dev version of tools (`pip install --upgrade --force-reinstall git+https://github.com/nf-core/tools.git@dev`) you can run `nf-core modules container create` to have them build and returned with the right syntax.

mashehu (approved):

dialvarezs (approved):
Looks good!

workflows/genomeassembler.nf
nschan: This is a good point, I added it because without it there were `null` entries in the list, even though I also would have expected them to be removed?

dialvarezs: Maybe because the nulls come inside a tuple, so, that's why the first one  doesnt work.
In that case  the last one is the correct way.

nschan:
I will merge this after #203,  which is fixing the test, and then the tests here should also pass again.

nschan:
@nf-core-bot fix linting

PR #207, opened 2026-07-21
More tests

nschan:
This PR adds additional tests for BUSCO and QUAST reporting.
Addition of these tests has uncovered some problems:

* QUAST output reader expected reference, this is now more flexible.
* BUSCO occasionally breaks when files have extension `fasta` https://gitlab.com/ezlab/busco/-/work_items/793, patched modules to emit everything with `.fa`
* BUSCO reader was adapted to changes in BUSCO output format

Test snapshots were (re-)generated


modules/nf-core/flye/main.nf
```diff
@@ -12,7 +12,7 @@ process FLYE {
     val mode
 
     output:
-    tuple val(meta), path("*.fasta")   , emit: fasta
+    tuple val(meta), path("*.fa")   , emit: fasta
```
Daniel-VM, line 15: I guess flye would need a patch as well?

modules/nf-core/flye/main.nf
Daniel-VM: Since the expected FASTA extension has changed, consider updating it in the stub section as well

modules/nf-core/pilon/main.nf
Daniel-VM: missing update

conf/modules/assembly.config
```diff
@@ -19,7 +19,7 @@ process {
         publishDir = [
             path: { "${params.outdir}/${meta.id}/assembly/flye/" },
             mode: params.publish_dir_mode,
-            saveAs: { filename -> filename.matches("*.assembly.fasta") ? null : filename }
+            saveAs: { filename -> filename.matches(".*.assembly.fa") ? null : filename }
```
Daniel-VM, line 22: You can leave the saveAs statement like this:

saveAs: { filename -> filename.equals('versions.yml') ? null : filename }

Then add a pattern key to define what you want to copy to the results directory. You can see an example [here](https://github.com/nf-core/bacass/blob/2c0626ebb0256e915f3962487c3e062a3ddbdb39/conf/modules.config#L408).

modules/nf-core/ragtag/patch/main.nf
Daniel-VM: Missing update?

Daniel-VM (commented):
I’ve added a few comments.

Maybe we could also consider a more targeted workaround by normalizing the extension only at the BUSCO module level, instead of changing the extension across multiple related modules. Have you already tested that option?

Could you please double-check the `stub` sections where the FASTA extension also needs to be changed?

conf/modules/assembly.config
nschan: This pipeline is wholly committed to the versions topic, and does not produce `versions.yml` in any place. This here is to avoid publishing the unzipped `.fa` file, since I bgzip all assemblies in an explicit process, as some of the nf-core modules produce `gzipped` files, incompatible with samtools.

nschan:
> Maybe we could also consider a more targeted workaround by normalizing the extension only at the BUSCO module level, instead of changing the extension across multiple related modules. Have you already tested that option?

I am not sure if that is more convenient in the end and I somehow prefer to have a unified extension scheme within this pipeline. For BUSCO, simple workarounds (e.g. `stageAs("input.fa.gz")`) will cause problems because the reporting module figures out which step of the pipeline was run through busco based on the filename. Maybe you have some other way in mind?

> Could you please double-check the `stub` sections where the FASTA extension also needs to be changed?

Good point, I missed this during patching, I will update.

conf/modules/assembly.config
Daniel-VM: Oh, that’s true... I completely forgot about the version topic update. I still need to do my work in bacass.

modules/nf-core/flye/main.nf
nschan: Added the patchfile.

Daniel-VM (approved):
LGTM!

PR #212, opened 2026-08-07
bump medaka to 2.2.2, closes #204

### LINKED ISSUE #204: substitute medaka image with an update version
### Description of feature

hi,
it is possible to manually modify the pipeline so to use an update image of medaka?
In the current version of the pipeline the medaka's models available go up to 5.0, but the newer minKNOW basecaller uses the 5.2. There is any workaround that is not downgrade the basecall to  5.0 model?

Thank you.

nschan:
As suggested in #204 , this updates medaka to version `2.2.2`


prototaxites (approved):
LGTM :)

PR #213, opened 2026-08-07
Pre-release: Bump 2.0

nschan:
Version bump to v 2.0.0, ran `nf-core p lint --release` with no failures, but some warnings.


prototaxites (commented):
Do you need to fix the version number in test snapshots?

nschan:
> Do you need to fix the version number in test snapshots?

I dont think the pipeline version actually goes into any snapshot?

prototaxites:
Genomeassembler is missing the call to the softwareVersionsToYAML function from the template, which inserts a "Workflow" section in the output YAML with the pipeline version:

https://github.com/nf-core/genomeassembler/blob/f5481172a9e062a155cd53fe8392eeea5448812b/workflows/genomeassembler.nf#L149-L155

Should be: 

https://github.com/nf-core/tools/blob/eb2f709090f4054f45437c34049ea2068567c339/nf_core/pipeline-template/workflows/pipeline.nf#L77-L84

```
    def ch_collated_versions = softwareVersionsToYAML(ch_versions.mix(topic_versions.versions_file))
        .mix(topic_versions_string)
        .collectFile(
            storeDir: "${outdir}/pipeline_info",
            name: {% if is_nfcore %}'nf_core_'  + {% endif %} '{{ short_name }}_software_' {% if multiqc %} + 'mqc_' {% endif %} + 'versions.yml',
            sort: true,
            newLine: true
        )
```

nschan:
Thanks for noticing that @prototaxites. I was unaware that `sofwareVersionsToYAML` injects the pipeline & nxf version. I had to do some slight modifications to make it work with the downstream reporting, the only change is that `.collectFile()` does not go into `ch_collated_versions` as I need a channel.

ro-crate-metadata.json
```diff
@@ -338,20 +337,10 @@
...
-            "@id": "https://orcid.org/0000-0003-1675-0677",
-            "@type": "Person",
-            "email": "mahesh.binzer-panchal@nbis.se",
-            "name": "Mahesh Binzer-Panchal"
-        },
-        {
-            "@id": "https://orcid.org/0000-0002-7860-3560",
-            "@type": "Person",
-            "email": "mm49@sanger.ac.uk",
-            "name": "Matthieu Muffato"
```
prototaxites, line 354: Were Mahesh and Matthieu meant to be removed here?

nschan: This was intentional, but I am not sure about the politics within nf-core here (i.e. what are the consequences of being listed here, does this entail responsibilities for the pipeline?), or how exactly they ended up here in the first place, so I am happy to discuss it. 
@mahesh-panchal had initially set up the genomeassembler repo, but current genomeassembler was an independent development and I contributed it since no genomeassembler pipeline had materialized. 
For @muffato, it is less clear to me. If there are contributions that were made that I forgot about, I apologize.

mahesh-panchal: It's likely Matthieu approved the first PR I made, and that's why he's included.

nschan: Should both of you remain included here?

FriederikeHanssen: Hey! We normally never remove anyone that has contributed. Their role might change from author to maintainer and vice versa, but their work still is part of the history that others build upon: 

https://nf-co.re/docs/contributing/contributor-types?utm_source=chatgpt.com#contributor
https://nf-co.re/docs/contributing/contributor-types?utm_source=chatgpt.com#best-practices
`Contributors can be both current and past.`
`Never remove anyone from the contributor list. Acknowledge everyone who has contributed to the pipeline at any point in its history, whether in version 1 or version 10.`

nschan: Ok, I will add them back in.

prototaxites: I would say rather that at a minimum, everyone who has contributed a fix via PR should be included in the manifest as a contributor? Currently this only includes @nschan.

https://github.com/nf-core/genomeassembler/pulls?q=is%3Apr+-author%3Anschan+

nschan: I do not have strong feelings either way. But maybe this issue comes up again in other repos.
My impression was that this should reflect contributions to the _pipeline_, which is why I updated this field. If this is about historical contributions to the _repo_, even if they never materialized into anything relevant for the current pipeline, I am also ok with that.

nschan: I would like to clarify this, because in my personal view @prototaxites made more contributions to this pipeline than @muffato, via reviews and discussions, and if the contributor section should reflect that (and not only direct code-contributions) I would update it.

muffato: Hi @nschan . You're right about wanting to add @prototaxites . Contributors are not restricted to code. Everyone who commenting on or reviewed a pull-request should have the `contributor` role at least. The list shouldn't have just you.

nschan: Great, thanks for clarifying, I will then add everyone who reviewed or otherwise contributed to PRs.

prototaxites (approved):
Looks good! Fine with me now, assuming tests pass.

ro-crate-metadata.json
nschan: I have now updated the manifest, and re-generated the ro-crate. This now includes (not tagging everyone to avoid pings):

  - People who have provided test data, so far uncredited: jkh00 , OlivierCoen 
  - People who contributed by reviewing: prototaxites , mirpedrol , mahesh-panchal, vagkaratzas , erikrikarddaniel, muffato, maxulysse , nvnieuwk, jffy113, louislenezet

While I tried to actually find information on most of the contributors, I noticed that many pipelines seem to exclusively list code contributors, and not PR reviewers etc. It appears that the rules for inclusion as a contributor are not very well established within nf-core.

Issue #214, opened 2026-08-27
Additional output: Input samplesheet for `nf-core/genomeqc`


nschan:
### Description of feature

As briefly discussed via [slack](https://nfcore.slack.com/archives/C07LY51P4S2/p1787560519894009), it would be nice to provide an easy way to chain `nf-core/genomeassembler` to `nf-core/genomeqc`. This would require a sample-sheet output:

```csv
species,fasta,gff
samplename,/path/to/fasta,/path/to/gff
```

where gff is optional.

The main question is how to deal with the intermediate assembly stages (which I think are valuable) in this, presumably the easiest way is to determine this is based on the [output directory structure](https://github.com/nf-core/genomeassembler/blob/dev/docs/output.md).

I am not sure if this would require a dedicated process, or should be done via nextflow/groovy. Presumably doing this within nextflow is more reasonable.

PR #215, opened 2026-08-28
Generate Genomeqc samplesheet

nschan:
Given the recent release of nf-core/genomeqc, preparation of a samplesheet for downstream genomeqc was proposed (#214).

This PR adds an extra output, compatible with the (updated) downstream samplesheet in the form of `assembly,fasta`.

Trying to implement this has uncovered a bunch of inconsistencies, which are fixed here:

 - The placeholder channel that contains all final scaffolds (there is nothing downstream currently), was misformatted and contained a nested map. This has been fixed in `subworkflows/scaffold/main.nf`. There was also a problem in one of the nested scaffolding workflows (`ragtag`) which did not emit the correct element

 - The local `BGZIP` process, which ensures that all assemblies are bgzipped for samtools compatibility emits a folder (`out/file.gz`), I had not realized that this will publish the same folder structure, and not simply the file. Fixed via fixes to `config/` files `publishDir` directives

`usage.md` was updated

`CHANGELOG.md` was updated

Test snapshots were updated.


workflows/genomeassembler.nf
```diff
@@ -231,6 +231,66 @@ workflow GENOMEASSEMBLER {
...
+                [ meta.id, 'polish_pilon', meta.polished ? meta.polished.pilon ?: null : null,            'polish/pilon'        ],
+                [ meta.id, 'polish_medaka', meta.polished ? meta.polished.medaka ?: null : null,           'polish/medaka'       ],
+                [ meta.id, 'polish_dorado', meta.polished ? meta.polished.dorado ?: null : null,           'polish/dorado'       ],
+                [ meta.id, 'initial_assembly', meta.assembly, subout
+                ]
+                ]
+        }
+        .flatMap( { it -> it } )
+        .filter {
+            _id, _stage,  file, _subdir -> file != null
```
erikrikarddaniel: Minor: naming this closure param `file` shadows Nextflow's built-in `file()` function for the rest of this closure's body. Harmless here since `file()` isn't called inside it, but easy to trip over later if someone adds a check that needs it.

```suggestion
            _id, _stage, assembly_file, _subdir -> assembly_file != null
```

erikrikarddaniel (approved):
Tiny suggestion, otherwise great.

workflows/genomeassembler.nf
nschan: Great catch, thanks a lot!

PR #217, opened 2026-09-02
Release linting

nschan:
Final (?) PR to dev before release PR. 

`nf-core p lint --release` noted a bunch of modules which received updates, I updated all of them, applied patches where necessary (input shape mostly). For `picard` modules, the input shape in the swf has been adjusted.

I also misunderstood nf-test test specifications, which resulted in the same test running 3 times, with different names. While I tested the actual test profiles in local tests, snapshots were identical for all tests..
I further realized that there was an issue that affected qc with `busco` if qc with `quast` was not running. This has been fixed, but the tests including busco take a substantial amount of disk space, and time. It remains to be seen if those are compatible with CI.


nschan:
It appears that `quast`, and the reporting module more broadly, create unstable files. I will investigate.

modules.json
```diff
@@ -61,72 +60,72 @@
...
-                        "git_sha": "22edd7ce9462025093b25e3b55006f7720ff0d8e",
+                        "git_sha": "37f69c8ca3eb3d038a5d44f33a329e49f44bd4f2",
                         "installed_by": ["modules"]
                     },
                     "jellyfish/count": {
                         "branch": "master",
-                        "git_sha": "614004b6a8e752f9f0d362e4fb94d67bb7b7455c",
+                        "git_sha": "f783ab0f09a75b5134084aa379280aef197b6dd8",
                         "installed_by": ["modules"],
                         "patch": "modules/nf-core/jellyfish/count/jellyfish-count.diff"
```
erikrikarddaniel: The `jellyfish-count.diff` patch file this references was deleted in this PR (its gzip-decompression logic looks like it got upstreamed into the official module's `main.nf` instead — the new `main.nf` already has equivalent `is_compressed`/`fasta_name` handling), but the `"patch"` key here wasn't removed to match. Next `nf-core modules update`/`nf-core modules patch` run (or `nf-core pipelines lint`'s patch check) will trip over a missing patch file.

```suggestion
                        "installed_by": ["modules"]
```

subworkflows/local/qc/main.nf
```diff
@@ -66,7 +66,20 @@ workflow QC {
...
+                }
+                .join(scaffolds)
+                .map {
+                    _id, meta, target_scaffolds ->
+                    [
+                      meta: meta + [qc_target: target_scaffolds] // QC Target only exists in QC channel, and takes the scaffold that should be qc'ed
+                    ]
+                }
+        )
+    ch_qc.dump(tag: "CH_QC after mapping:")
```
erikrikarddaniel, line 82: Leftover debug `.dump()` from tracking down the busco/quast bug — harmless (only fires with `-dump-channels`) but probably shouldn't ship in the release PR.

```suggestion
```

subworkflows/local/qc/main.nf
```diff
@@ -97,6 +110,8 @@ workflow QC {
                 busco_db: it.meta.busco_db ? file(it.meta.busco_db, checkIfExists: true) : []
             }
 
+    busco_in.fasta.dump(tag: "QC: BUSCO FASTA: ")
```
erikrikarddaniel, line 113: Same as the other `.dump()` above — leftover debugging output.

```suggestion
```

tests/default.nf.test
```diff
@@ -67,11 +68,16 @@ nextflow_pipeline {
             )
         }
     }
+    /*
```
erikrikarddaniel: Both `test_busco` and `test_busco_quast` — the only nf-test cases exercising BUSCO, including the busco-without-quast combination that was the actual bug just fixed in this PR — are now fully commented out for CI cost reasons. That leaves the regression this PR fixes with no CI coverage going forward. Since the PR description already flags this as unresolved ("It remains to be seen if those are compatible with CI"), might be worth settling one way or the other before merge — e.g. a lighter-weight busco fixture, or explicitly accepting the coverage gap rather than leaving it as a commented-out block.

erikrikarddaniel (approved):
You have a left-over patch in `modules.json` that's important to remove. I'm not sure Claude got the suggestion right there; it looks like `"installed_by"` will be repeated.

A few other small comments, otherwise good to go.

Approving despite the `modules.json` since it's your pipeline.

subworkflows/local/qc/main.nf
nschan: There are more dump tags throughout my pipeline. My understanding was that since these needs `-dump-channels` explicitly (unlike `.view{}`) they are ok?

tests/default.nf.test
nschan: Yup. I will remove those tests here. I for now accept that this is not CI tested.

nschan:
Looks like there is something more thats unstable..

subworkflows/local/qc/main.nf
erikrikarddaniel: You're right — I only checked this file in isolation and missed that `.dump()` is already used throughout the pipeline (`assemble/main.nf`, `prepare/*`, `scaffold/*`, etc.). Since it needs `-dump-channels` explicitly to do anything, it's consistent with the existing pattern here — no need to remove it. Sorry for the noise.

subworkflows/local/qc/main.nf
erikrikarddaniel: You're right — I only checked this file in isolation and missed that `.dump()` is already used throughout the pipeline (`assemble/main.nf`, `prepare/*`, `scaffold/*`, etc.). Since it needs `-dump-channels` explicitly to do anything, it's consistent with the existing pattern here — no need to remove it. Sorry for the noise.

Issue #220, opened 2026-09-09
String paths instead of Nextflow staged Path objects


ewels:
### Description of the bug

QUAST fails in the QC subworkflow whenever a reference is used (`use_ref`), with:

```
ERROR! File not found (reference_bam): .../<id>_to_reference.bam
```

**Root cause:** the reference-mapping BAM and assembly-mapping BAM are passed to QUAST as **raw path strings embedded in `meta`** and injected via `ext.args`, rather than being declared as `path` inputs on the process. Because they are not declared inputs, Nextflow never stages them into the QUAST task work directory, so the interpolated path points at the *producer's* work directory, which QUAST cannot resolve in its own isolated task filesystem. This breaks on cloud/Fusion execution (AWS Batch) and any executor where task work dirs are isolated.

Trace of the mechanism:

1. `subworkflows/local/mapping/map_to_ref/main.nf` emits `ch_aln_to_ref_bam = [meta, bam]` (`<id>_to_reference.bam`).
2. `subworkflows/local/assemble/main.nf` stuffs the BAM path into `meta`:
   ```groovy
   ch_main_to_qc = MAP_TO_REF.out.ch_aln_to_ref_bam
       .map { meta, bam -> [ meta: meta + [ref_map_bam: bam] ] }
   ```
3. `conf/modules/QC/quast.config` injects it as a string (all 7 `withName` blocks):
   ```groovy
   ext.args = {
       [
           meta.ref_map_bam      ? "--ref-bam ${meta.ref_map_bam}" : '',
           meta.assembly_map_bam ? "--bam ${meta.assembly_map_bam}" : '',
           '--glimmer',
           '--large'
       ].join(" ")
   }
   ```
4. `modules/nf-core/quast/main.nf` declares **no BAM inputs**:
   ```groovy
   input:
   tuple val(meta) , path(consensus)
   tuple val(meta2), path(fasta)
   tuple val(meta3), path(gff)
   ```

The `.bai` indexes QUAST needs are likewise never staged.

### Suggested fix

Declare the BAMs (and their indexes) as proper optional `path` inputs on the QUAST module, wire them through `subworkflows/local/qc/main.nf` (via the existing `multiMap`), and drop the `--ref-bam`/`--bam` interpolation from `conf/modules/QC/quast.config` (keeping only `--glimmer --large`). Since `quast` lives under `modules/nf-core/`, apply this via `nf-core modules patch` or convert to a local module so it survives updates.

### Command used and terminal output

```console
[...]

PR #221, opened 2026-09-09
Pipeline object storage fix

nschan:
This fixes a bug that was discovered during AWS testing, which revealed that the way that some files were conditionally smuggled into quast, via `ext.args`, fails with object storage. The quast module has been patched to take those as inputs, QC subworkflow now passes those as inputs. The files are no longer passed via config.

Edit: reported in #220

### DIFF
--- a/conf/modules/QC/quast.config
+++ b/conf/modules/QC/quast.config
@@ -3,8 +3,6 @@ process {
         ext.prefix = { "${meta.id}_assembly" }
         ext.args = {
             [
-                    meta.ref_map_bam ? "--ref-bam ${meta.ref_map_bam}" : '',
-                    meta.assembly_map_bam ? "--bam ${meta.assembly_map_bam}": '',
                     '--glimmer',
                     '--large'
             ]
@@ -19,8 +17,6 @@ process {
         ext.prefix = { "${meta.id}_pilon" }
         ext.args = {
             [
-                    meta.ref_map_bam ? "--ref-bam ${meta.ref_map_bam}" : '',
-                    meta.assembly_map_bam ? "--bam ${meta.assembly_map_bam}": '',
                     '--glimmer',
                     '--large'
             ]
@@ -35,8 +31,6 @@ process {
         ext.prefix = { "${meta.id}_medaka" }
         ext.args = {
             [
-                    meta.ref_map_bam ? "--ref-bam ${meta.ref_map_bam}" : '',
-                    meta.assembly_map_bam ? "--bam ${meta.assembly_map_bam}": '',
                     '--glimmer',
                     '--large'
             ]
@@ -51,8 +45,6 @@ process {
         ext.prefix = { "${meta.id}_links" }
         ext.args = {
             [
-                    meta.ref_map_bam ? "--ref-bam ${meta.ref_map_bam}" : '',
-                    meta.assembly_map_bam ? "--bam ${meta.assembly_map_bam}": '',
                     '--glimmer',
                     '--large'
             ]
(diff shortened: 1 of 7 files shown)


nschan:
While looking at the configs I realized that I never took care of handling the QC steps for dorado (not fully supported, experimental).

conf/modules/QC/quast.config
```diff
@@ -35,8 +31,20 @@ process {
...
+                    '--large'
+            ]
+            .join(" ")
+        }
+        publishDir = [
+            path: { "${params.outdir}/${meta.id}/QC/QUAST" },
+            mode: params.publish_dir_mode
+        ]
+    }
+    withName: '.*DORADO:.*:QUAST' {
```
erikrikarddaniel, line 44: This DORADO block (and its counterparts in `alignments.config`/`busco.config`) isn't mentioned in the PR description, which only talks about the object-storage fix. It looks like a legitimate second fix riding along: `subworkflows/local/polishing/dorado/main.nf` already calls the shared `QC` subworkflow on `dev`, but none of these three config files had a `.*DORADO:...` selector before this PR, and none of them have a generic fallback block either -- so DORADO-polished assemblies' QUAST/BUSCO/alignment output would previously have had no `ext.prefix`/`publishDir` at all (i.e. not published). Worth a `CHANGELOG.md` line for this, since it's a distinct fix from the one stated in the PR description.

erikrikarddaniel (approved):
Reviewed with some help from Claude Code.

Traced the actual fix mechanism, not just the diff: `assembly_map_bam`/`ref_map_bam` are real Nextflow `Path` objects that `QC`'s own logic stashes into the `meta` map (e.g. `meta + [ assembly_map_bam: assembly_map_bam ]` from `MAP_TO_ASSEMBLY.out.aln_to_assembly_bam`). Before this PR they were only ever pulled back out via `${meta.assembly_map_bam}` string interpolation inside `ext.args` -- Nextflow never sees that as a file to stage, so under object storage the string that lands in the shell command is a remote URI the container can't read locally. Declaring `ref_map_bam`/`assembly_map_bam` as real `tuple val(meta), path(...)` process inputs (matching how `fasta`/`gff` already work as `meta2`/`meta3`) fixes exactly that: Nextflow now stages/localizes the file before the task runs, storage-backend-agnostic. This is the right fix for the described bug class, and the `[] `-as-"no file" idiom used in `subworkflows/local/qc/main.nf` (`it.meta.ref_map_bam ?: []`) matches the existing `use_ref`/`use_gff` pattern exactly.

Three non-blocking things, not blockers given this is a maintainer-authored fix with green CI (one as an inline comment below):

1. Since the original bug (#220) is specifically about object storage, and this repo's CI only runs against local disk (the `docker` profile jobs), CI passing here can't actually prove the fix works on the environment where the bug was first observed. Worth a quick manual confirmation on S3 (or wherever #220 was seen) if that hasn't already happened outside CI.
2. `modules/nf-core/quast/meta.yml` still documents only the original 3 inputs (`meta`/`consensus`, `meta2`/`fasta`, `meta3`/`gff`) -- it's now out of sync with the module's actual 5-input signature, since `ref_map_bam`/`assembly_map_bam` (`meta4`/`meta5`) aren't documented there.
3. `modules/nf-core/quast/tests/main.nf.test` wasn't updated (confirmed via the `.diff` file's own "unchanged" note) -- none of its 3 test cases set `input[3]`/`input[4]`, so the new `--ref-bam`/`--bam` argument-construction code this PR adds to the module's script block has no direct test coverage. CI staying green just means the existing cases still pass with those two inputs implicitly unset, not that the new logic was exercised.

Approving with these as comments -- nothing here blocks merging.

nschan:
Thanks. 

This will be tested via aws full test, which is also how this was discovered. I don't have an object storage setup where i could test it.

I do not think that it is required to update module tests for pipeline specific patches, since as I understand thw module test is mainly to ensure integrity of modules in the modules repo, while pipelines undergo pipeline tests.

Finally, since dorado was not part of any previous release, I do not see a point in adding this to changelog, as it fixes an issue that was never published. I acknowledge that it would have been cleaner to have this change in a separate pr, though.

PR #223, opened 2026-09-11
retry LINKS if exitcode == 2

nschan:
This PR adds exitcode 2 as a retry code for LINKS.

During AWS full test, this error occured:

```
Caused by:
  Essential container in task exited - OutOfMemoryError: Container killed due to memory usage
``` 

However, LINKS appears to not report this with an exitcode that indicates OOM, instead, LINKS exits with `2`:

```
Command exit status:
  2
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,10 @@ Pull requests in reverse chronological order since v1.1.0
 
 - Fixed typo in medaka url (@TomHarrop)
 
+[#223](https://github.com/nf-core/genomeassembler/issues/223)
+
+- Add exitcode `2` as retry code for links in AWS fulltest profile.
+
 ### `Dependencies`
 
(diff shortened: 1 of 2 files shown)


Joon-Klaps (approved):
If it's an OOM, I wonder if just retrying will fix this. 

Might be better then to allocate more memory to it with a `label 'process_high'`, or is that what you are trying to achieve with retry as `memory = { X.GB * task.attempt }` is somewhere set? 

Also changelog is not patched. 

But given that these comments are nitpicks, I'll approve

nschan:
> Might be better then to allocate more memory to it with a label 'process_high', or is that what you are trying to achieve with retry as memory = { X.GB * task.attempt } is somewhere set?

Yes, the idea is to retry with more memory, using the standard approach from `base.config`.

PR #226, opened 2026-09-11
fix buggy meta-map in scaffold outputs

nschan:
AWS fulltest failed, again. This time with
```
Cannot get property 'strategy' on null object
```

With seqera AI, this was traced to the `scaffold` swf, an issue which should have been fixed before, but was only fixed for the first couple of cases [#215](https://github.com/nf-core/genomeassembler/pull/215/changes/f181bdf94e7c97b572568bc41d05b6059a9c2997) 

Since the channel was not a map, `meta.strategy` selection in did not work for those cases (cp: [this commit in #215](https://github.com/nf-core/genomeassembler/pull/215/changes/cba85c3f5efa61326f740caff202743e2693b487).

### DIFF
--- a/subworkflows/local/scaffold/main.nf
+++ b/subworkflows/local/scaffold/main.nf
@@ -92,9 +92,16 @@ workflow SCAFFOLD {
                 )
                 .map {
                     _id, meta_links, meta_longstitch ->
-                        meta_links -
-                         meta_links.subMap("scaffolds_links") +
-                         [scaffolds: [links: meta_links.scaffolds_links, longstitch: meta_longstitch.scaffolds_longstitch]]
+                    [
+                        meta:   meta_links -
+                                meta_links.subMap("scaffolds_links") +
+                                [
+                                    scaffolds: [
+                                        links: meta_links.scaffolds_links,
+                                        longstitch: meta_longstitch.scaffolds_longstitch
+                                        ]
+                                ]
+                    ]
                 }
         )
         //links-ragtag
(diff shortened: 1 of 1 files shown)


erikrikarddaniel (approved):
Reviewed with some help from Claude Code.

The fix is correct. SCAFFOLD's single-scaffold-tool branches (and hic_out) already wrap their output as [meta: ...] — with a comment right above noting "the individual sswfs return map-channels, this means that it -> it.meta needs to be used, otherwise there is a nesting", added in https://github.com/nf-core/genomeassembler/pull/215. The four combination branches this PR touches (links+longstitch, links+ragtag, longstitch+ragtag, and the triple-combo) were still emitting a bare meta map instead of [meta: ...] — exactly the same bug https://github.com/nf-core/genomeassembler/pull/215 fixed elsewhere, just missed for the multi-tool-combination cases. SCAFFOLD.out.ch_main flows into workflows/genomeassembler.nf and is read as it.meta.id/.strategy downstream, which matches the Cannot get property 'strategy' on null object error in the PR description. The diff is purely mechanical — same wrapping pattern applied consistently, no other logic changed.

One non-blocking observation: there's no subworkflows/local/scaffold/tests/, so nothing in the regular CI (all green here) actually exercises the multi-tool-strategy combination paths — that's why this needed a full-scale run to surface, twice now (once for the single-tool case via https://github.com/nf-core/genomeassembler/pull/215, now for the combination cases). Worth a follow-up issue for subworkflow-level nf-test coverage of the double/triple-scaffold-tool paths, but not something to hold this fix on.

nschan:
Complete correct on the lack of testing for the later stages, of the pipeline. This is largely due to CI constraints. But, as tests are the best way to identify whats broken, I should figure out how to improve CI testing. Currently, my focus is finally getting this released. I will open an issue as a reminder.

PR #230, opened 2026-09-17
adjust bwamem2 index memory request for compressed input

nschan:
`bwamem2_index` is requesting too little memory with a `bgzipped` input and OOMs on AWS. I assume that this because the module calculates required memory based on filezise, which probably approximates sequence length fine for uncompressed files, but appears to undererstimate for compressed files.


modules/nf-core/bwamem2/index/main.nf
```diff
@@ -2,7 +2,8 @@ process BWAMEM2_INDEX {
     tag "$fasta"
     // NOTE Requires 28N GB memory where N is the size of the reference sequence, floor of 280M
     // source: https://github.com/bwa-mem2/bwa-mem2/issues/9
-    memory { 280.MB * Math.ceil(fasta.size() / 10000000) * task.attempt }
+    // patch for bgzipped input by assuming a 3x compression.
+    memory { 3 * 280.MB * Math.ceil(fasta.size() / 10000000) * task.attempt }
```
mahesh-panchal, line 6: Maybe add to the comment this assumes input is always bgzipped.

mahesh-panchal (approved):
Don't forget the change log entry

modules/nf-core/bwamem2/index/main.nf
nschan: Expanded the comment to clarify that the pipeline handles all fastas in bgzipped format.

